Issue #23, opened 2015-12-17
install Sensor on Centos?


ubun3t:
Hi, I am trying to install a sensor on Centos 6? Is possible? 

Thanks in advance!

stamparm:
There should be no problems. Only real requirement is pcapy (`sudo yum install pcapy`)

ubun3t:
Hi, sorry for delay! That's right. sensor's are working !! Thanks for your fast reply and your work!!!

PR #33, opened 2015-12-22
Added distro specific package requirements

jayzcode:
Based on query in issue #23.
The warning and errors messages must be distro specific in the benefit of the consumer. 
In this patch I have added a distro check and added the corresponding action.

### DIFF
--- a/sensor.py
+++ b/sensor.py
@@ -25,6 +25,7 @@ import time
 import traceback
 import urllib
 import urlparse
+import platform
 
 from core.common import check_sudo
 from core.common import load_trails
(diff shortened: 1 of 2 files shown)


stamparm:
Just a minor observation :)

`if (platform.linux_distribution()[0].lower() == 'fedora' or 'centos')` will always result with `True` :)

p.s. will merge and patch

stamparm:
Fixed.

@jayzcode thx for contribution ;)

jayzcode:
@stamparm I really liked the idea. Happy to help..!!

PR #93, opened 2016-06-30
Implementation for an Issue #4

laciKE:
Support for selection of multiple days.

Use mouseclick in calendar (heatmap) with CTRL-key pressed for selection of the first and the last day.

### DIFF
--- a/core/httpd.py
+++ b/core/httpd.py
@@ -394,10 +394,38 @@ def start_httpd(address=None, port=None, join=False, pem=None):
 
             start, end, size, total = None, None, -1, None
             content = None
-            event_log_path = os.path.join(config.LOG_DIR, "%s.log" % params.get("date", ""))
+            log_exists = False
+            dates = params.get("date", "")
+            if dates.find("_") == -1:
+                event_log_path = os.path.join(config.LOG_DIR, "%s.log" % dates)
+                if os.path.exists(event_log_path):
+                    range_handle = open(event_log_path, "rb")
+                    log_exists = True
+            else:
+                if dates.count("_") == 1:
+                    logs_data = ""
+                    date_interval = dates.split("_")
+                    try:
+                        start_date = datetime.datetime.strptime(date_interval[0], "%Y-%m-%d").date()
+                        end_date = datetime.datetime.strptime(date_interval[1], "%Y-%m-%d").date()
+                        for i in range(int((end_date - start_date).days) + 1):
+                            date = start_date + datetime.timedelta(i)
+                            event_log_path = os.path.join(config.LOG_DIR, "%s.log" % date.strftime("%Y-%m-%d"))
+                            if os.path.exists(event_log_path):
+                                log_handle = open(event_log_path, "rb")
+                                logs_data += log_handle.read()
+                                log_handle.close()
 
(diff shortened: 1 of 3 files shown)


stamparm:
@laciKE will take a look on Monday :)

stamparm:
Thank you :)

p.s. still have to skip the `logs_data += log_handle.read()` and work with streams (because, sometimes log files can be huge)

jdrews:
Hmm strange. I just tried this and it's not working. CTRL + mouse click on calendar lets me select one day after another (red square on only one day). But nothing is updated after that...

I confirmed it's not a cache issue and I confirmed I'm running with your changes...

laciKE:
@stamparm thanks for merge and comment, I will try to resolve a problem with huge log files. What about solution like this?
```
  if http_header.range, then
    foreach log_file use os.stat(log_file).st_size
    determine which log files are affected by range (start, end)
    read only logs_data from files mentioned above
  else
    use current approach (read logs_data from each log_file in requested interval)
```

@jdrews thanks for feedback, could you test following steps, pleas?
1) in browser go to the `[YourMaltrailURL]/events?date=2016-07-01_2016-07-04` (e.g `http://localhost:8338/events?date=2016-07-01_2016-07-04`). Do you see your logs from specified dates interval?
2) Open Developer tools or Browser Console in your browser (Chrome: Ctrl+Shift+I, Network tab; Firefox Ctrl+Shift+J, right click -> check Log Request and Response Bodies). Then select two days in calendar (hold CTRL-key all the time). Could you see requests like `events?date=2016-07-01_2016-07-04` in console? The response body of these requests should be similar than response in step 1.
@stamparm , do you have the same problem like @jdrews ? Or do you have any idea what caused this problem?

stamparm:
@laciKE nope. it worked pretty well at my side

jdrews:
@laciKE I can confirm that [YourMaltrailURL]/events?date=2016-07-01_2016-07-04 works. I see all logs from 07-01 to 07-04. 

I do not see any events in the Chrome debugger network tab that look like events?date=2016-07-01_2016-07-04 when holding CTRL and clicking on dates in the calendar/heatmap. 

I turned on some breakpoints and watched CTRL_DATES. I'm noticing that on the first CTRL + click I'm seeing the CTRL_DATES array populate with the first clicked day. But on the next CTRL + click this array is cleared out by the time line [261 ](https://github.com/laciKE/maltrail/blob/f28a516a6cec47f7590c4213cb0fd54f463dae42/html/js/main.js#L261)comes around. Then the next day gets added on [267](https://github.com/laciKE/maltrail/blob/f28a516a6cec47f7590c4213cb0fd54f463dae42/html/js/main.js#L267), but we never hit 2 days to trigger a repaint.

laciKE:
@jdrews thanks for your investigation. It is strange, because the only place, where the CTRL_DATES can be cleared is on line 84, when user is pressing CTRL key down. So just now I do not see any reason why this array is cleared without releasing and pressing CTRL key... 
Just for clarification, did you release the CTRL key between 1st and 2nd click on heatmap or did you press CTRL key all the time?

jdrews:
I pressed the CTRL key the entire time.

PR #110, opened 2016-11-01
add some feed source

th3gundy:
add threat intelligence sources


stamparm:
`http://malwareurls.joxeankoret.com/domains.txt` is dead. If you go there you'll see a notice:

```
# SERVICE SHUT-DOWN. THE SERVICE WILL NOT BE AVAILABLE AGAIN.
# Notice: http://malwareurls.joxeankoret.com
```

stamparm:
`http://data.phishtank.com/data/online-valid.json` is 50MB :D. Not really sure that it is worth for loading into the memory (as part of trails). Also, those URLs have a short life because web browsers automatically check for those URLs prior to visit

th3gundy:
http://malwareurls.joxeankoret.com/normal.txt also dead :)
I think phishtank.com is a good source for OSINT Threat Intelligence feed like openphish.com. But 50MB size, its very high like you say. Thats your decision

PR #119, opened 2017-01-06
DataPlane feeds added

jtkdpu:
The Dragon Research Group is dead.  Those feeds have not been updating since May 2016.  The web page for those links is likely going to go away eventually.  I was the managing director of that group, but it disbanded once I left Team Cymru.  However, a new project, DataPlane.org has some similar feeds that you may wish to use.  Based on the dragonresearchgroup feeds I've added the DataPlane feeds for your convenience.


stamparm:
I will "compress" those three SIP feeds scripts into one. Also, will change __info__ data to be compliant with other feeds (e.g. know attacker (SIP), known attacker (SSH)...). Ok?

jtkdpu:
No complaints from me.

stamparm:
@jtkdpu thank you for your contribution

PR #132, opened 2017-05-03
Enforce usage of urllib2 instead of urllib for updates

ptitdoc:
urllib1 SSL implementation is broken in some situations (for instance when using specific web proxies...).

### DIFF
--- a/core/update.py
+++ b/core/update.py
@@ -14,7 +14,7 @@ import sqlite3
 import subprocess
 import sys
 import time
-import urllib
+import urllib2
 import urlparse
 
 sys.dont_write_bytecode = True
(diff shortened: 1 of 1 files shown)


stamparm:
Can you please be more specific? In what kind of situations? Is there any reference for your claims? Also, SSL is broken in lots of situations for all Python libraries prior to 2.7.9/2.7.10

ptitdoc:
I discovered this bug several month ago caused by a Fortnet web proxy (I searching for the reference I found at that time).

ptitdoc:
https://docs.python.org/3.4/howto/urllib2.html:

> Currently urllib.request does not support fetching of https locations through a proxy. However, this can be enabled by extending urllib.request as shown in the recipe

ptitdoc:
Here is another reference (maybe a better one):
https://chromiumcodereview.appspot.com/10825107/

PR #137, opened 2017-06-13
simple ignore rule configured by file misc/ignore_event.txt

arons:
Ignore matching event with simple rules.

### DIFF
--- a/core/ignore.py
+++ b/core/ignore.py
@@ -27,32 +27,8 @@ import subprocess
 import urllib2
 import zipfile
 import zlib
-
-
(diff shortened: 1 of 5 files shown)


stamparm:
Makes sense. If I could only ask you to make a new configuration entry inside `maltrail.conf`, like `USER_IGNORELIST` (pretty similar to `USER_WHITELIST`), instead of looking into the existence of `misc/ignore_event.txt` - as it is better to put those user defined files outside of the installation directory

arons:
I've added it and moved reading methods into settings.py

PR #177, opened 2018-05-13
Create elf_mirai.txt

MikhailKasimov:
Added example of updatible file for Linux.Mirai IP-s static trails. This file should be as add-on to /maltrail/trails/feeds/ main public Linux.Mirai IP-s list (which is currently absent in Maltrail feeds-list), because some IP-s can be missed in main trail-list.

Address of public Linux.Mirai IP-s list: [0] http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt , which should compulsory added to /maltrail/trails/feeds/

Also addresses from [0] can be replicated to static elf_mirai.txt trails to prevent case, if [0] is unaccessible for some reason, and not to loose detection of Linux.Mirai trails.

Current IP-s, mentioned in elf_mirai.txt are absent in [0]. But detux.org returns couple of Linux.Mirai samples on them. That's why I consider this file (elf_mirai.txt) as an add-on to main feed-trail from [0], which is also should be added to Maltrail feeds.

Thanks!

### DIFF
--- a/trails/static/malware/elf_mirai.txt
+++ b/trails/static/malware/elf_mirai.txt
@@ -1,0 +1,10 @@
+# Copyright (c) 2014-2018 Miroslav Stampar (@stamparm)
+# See the file 'LICENSE' for copying permission
+
+# Reference: http://blog.malwaremustdie.org/2016/08/mmd-0056-2016-linuxmirai-just.html
+# Reference: https://www.fortinet.com/blog/industry-trends/iot-based-linux-mirai-frequently-asked-questions.html
+
+89.186.189.241
+142.58.24.170
+143.239.27.171
+182.232.250.171


MikhailKasimov:
maltrail_ 2018-05-13 _-_2018-05-13_08 09 20

stamparm:
IPs are highly undesirable IOCs, especially if coming from report dated 2016

MikhailKasimov:
> IPs are highly undesirable IOCs, especially if coming from report dated 2016

http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt
```
# 31782018f2fc5d7a888a19f9e68a19bd
# CNC BOTNET IP BLOCKLIST
# http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt
# updated Mon May 14 06:27:06 UTC 2018   <------------------   
```

stamparm:
Not sure what http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt has to do with my statement `IPs are highly undesirable IOCs, especially if coming from report dated 2016`. You've pushed 4 IPs from 2016 report and now you are pasting here a totally different IP list.

Also, I went through that new list and I would say that it is falsely stating that it carries Mirai's "CNC BOTNET IPs". It maybe carries up to date IPs of Mirai infected machines, though, those same machines are something completely different from CNCs. When capturing in-network Malware you would use something that goes OUTBOUND as an indicator (e.g. resolving of a domain name or fresh-24h CNC IP list), while ignoring INBOUND traffic from infected IP addressed. Also, those same IPs from that same list (http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt) can be found in numerous lists that are already being pulled (e.g. ciarmy) in Maltrail.

MikhailKasimov:
> You've pushed 4 IPs from 2016 report and now you are pasting here a totally different IP list.

It was initially mentioned ``` Address of public Linux.Mirai IP-s list: [0] http://sanyalnet-cloud-vps.freeddns.org/mirai-ips.txt , ```

Other arguments are gotten. Thanks for clarification! :smile:

PR #178, opened 2018-05-13
Create elf_slexec.txt

MikhailKasimov:
From [0]: http://blog.malwaremustdie.org/2015/07/mmd-0037-2015-bad-shellshock.html

"In the first two samples we grabbed, it kills the listed process name which was downloaded from the below hostname AND IP:

kill.et2046.com 
sb.et2046.com
115.23.172.31"


stamparm:
Will merge, but going to remove IP (report is from 2015)

PR #179, opened 2018-05-13
Update elf_xorddos.txt

MikhailKasimov:
Added IPs from [0] http://blog.malwaremustdie.org/2015/07/mmd-0037-2015-bad-shellshock.html


stamparm:
Same as in other comments.

p.s. going to remove that old IP from previous list

PR #183, opened 2018-05-14
Create nigelthorn.txt

MikhailKasimov:
IoC from [0] https://blog.radware.com/security/2018/05/nigelthorn-malware-abuses-chrome-extensions/

### DIFF
--- a/trails/static/malware/nigelthorn.txt
+++ b/trails/static/malware/nigelthorn.txt
@@ -23,3 +23,11 @@ mxoonlites.com
 cyank.com
 lnlinvdeoa.com
 soeqpai.com
+
+# JS Miner Domains
+redirect35.info
+redirect39.info
+redirect41.info
+redirect44.info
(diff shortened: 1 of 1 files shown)


stamparm:
Going to accept only C2s. "Infected Domains" and "JS Miner Domains" are not reliable indicators. They are either expected only once (which is not proof of infection) or some other JS mining instance could be run at some other page (again not proof of infection)

MikhailKasimov:
```DNS (domain names) are desirable trails in malware cases``` (from 182) <--> ```"Infected Domains" ... are not reliable indicators.```

:confused:

stamparm:
Infected domains are not C2s. Infected domains are something used for spreading malware. Those kind of cases are cleaned FAST and are not reliable way of detecting malware (only possible infection in time of carrying the malicious scripts)

MikhailKasimov:
OK.

PR #184, opened 2018-05-14
Update whitelist.txt

MikhailKasimov:
::1 -- IPv6 localhost added.

1.1.1.1 is Cloudflare public DNS. Added in PR #170. Remove Bogus group.

symcd.com -- added to "triggering suspicious http request" group.


MikhailKasimov:
Any change to be considered?

PR #212, opened 2018-08-03
Update crypto_mining.txt

MikhailKasimov:
- Added site-list from:

[0] https://www.proofpoint.com/us/threat-insight/post/if-you-cant-steal-them-mine-them-cryptocurrency-threat-roundup


stamparm:
Those are addresses of valid crypto-currency exchanges. They are not related to the crypto-mining whatsoever.

Regular users will try to access those (personally, I am against marking those attempts as malicious and/or suspicious).

MikhailKasimov:
OK, thanks!

PR #216, opened 2018-08-07
Update ramnit.txt

MikhailKasimov:
[0] https://research.checkpoint.com/ramnits-network-proxy-servers/

Addressed from Figure 3.

### DIFF
--- a/trails/static/malware/ramnit.txt
+++ b/trails/static/malware/ramnit.txt
@@ -28,5 +28,4 @@ ramilhgme.com
 havonolwc.com
 anulwyqw.com
 mtankfqv.com
-zahlung.name
 ngioweb.su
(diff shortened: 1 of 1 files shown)


stamparm:
p.s. just noticed that those are already covered with https://github.com/stamparm/maltrail/blob/master/trails/feeds/bambenekconsultingc2dns.py, but nevertheless, nice to have it as static trails

MikhailKasimov:
BTW, there are also strong feed-lists, as it was for 36.com DGA-feeds, to add to MT: http://osint.bambenekconsulting.com/feeds/  Please, consider!

stamparm:
@MikhailKasimov out of all those, combined lists with "high confidence" are already being used ("High-Confidence C2 Domain Feed", "High-Confidence C2 IP Feed") along with "DGA Domain Feed". Hence, there is no need to add anything else (IMO)

MikhailKasimov:
OK, thnx.

PR #219, opened 2018-08-07
Create sinkhole_kryptoslogic.txt

MikhailKasimov:
See 2) of description from https://github.com/stamparm/maltrail/pull/218/


stamparm:
Sinkhole trails should be IPs (lots of different domains get sinked there; also, afterwards, if those IPs are not used anymore (rarely), they are not so popular to be transfered). Going to push a proper commit.

PR #230, opened 2018-08-17
Create firefox_banned_addons.txt

MikhailKasimov:
Hello!

[0] https://bugzilla.mozilla.org/show_bug.cgi?id=1483995
[1] https://www.bleepingcomputer.com/news/security/mozilla-removes-23-firefox-add-ons-that-snooped-on-users/

From [1]: ``` Besides Web Security, other banned add-ons include Browser Security, Browser Privacy, and Browser Safety. All of these have been observed sending data to the same server as Web Security, located at 136.243.163.73. ```

I remember, IPs are not acceptable in most cases for MT, but here it seems to be an exclusion to have the respective static trail. 

Please, consider. Thanks!

### DIFF
--- a/misc/firefox_banned_addons.txt
+++ b/misc/firefox_banned_addons.txt
@@ -1,0 +1,7 @@
+# Copyright (c) 2014-2018 Miroslav Stampar (@stamparm)
+# See the file 'LICENSE' for copying permission
+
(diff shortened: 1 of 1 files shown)


stamparm:
Will modify according to https://www.bleepingcomputer.com/news/security/firefox-add-on-with-220-000-installs-caught-collecting-users-browsing-history/ (to prevent false-positives)

PR #231, opened 2018-08-17
Create marap.txt

MikhailKasimov:
[0] https://www.proofpoint.com/us/threat-insight/post/new-modular-downloaders-fingerprint-systems-prepare-more-part-1-marap

```Marap C&C```. Unfortunately, here is nothing except IPs, which are workable. Later they can be moved to ```bad history``` trail, when they become unreachable.


stamparm:
Going to merge and adapt according to that reference (http://..../dot.php)

MikhailKasimov:
OK.

PR #245, opened 2018-08-23
Update lokibot.txt

MikhailKasimov:
[0] https://twitter.com/malwrhunterteam/status/1032537769787183104


stamparm:
p.s. in this kind of cases prefer usage of ONLY main domain (in this case `americaircairmakan.com`)

PR #257, opened 2018-08-24
Update lokibot.txt

MikhailKasimov:
- Added addresses from [0] https://twitter.com/FewAtoms/status/1033040103155871744 talk-flow.

Looking on ```occe.com``` and ```185.185.40.152```, we see the constant snippet ```fre.php```. Can we have some heuristic, based on this construction to detect ```lokibot```? **UPD**: Also ```PvqDq929BSx_A_D_M1n_a.php``` is applicable.


stamparm:
Already possible to use `/path` as as a trail, but not recommended as it could easily produce false positives

PR #261, opened 2018-08-28
Update lokibot.txt

MikhailKasimov:
Lokibot C2 address from [0] https://myonlinesecurity.co.uk/slightly-different-lokibot-delivery-via-embedded-ole-objects-in-rtf-word-doc/

### DIFF
--- a/trails/static/malware/lokibot.txt
+++ b/trails/static/malware/lokibot.txt
@@ -267,3 +267,7 @@ nawck.ml
 mitch-portal.tk
 sintrol.cf
 sirmitch.ml
+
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
Have rechecked on ```http://kikehraeein.com/web-obtain/file/web/log/Panel/five/fre.php``` -- no reaction from MT.

```Maltrail (server) #v0.10.373``` is in use.

@stamparm Take a look, pls.

stamparm:
@MikhailKasimov you used url with an upper letter `P`. It seems that I haven't lower-cased the path during the URL processing. With the latest commit this should be patched

PR #264, opened 2018-08-30
Create quasarrat.txt

MikhailKasimov:
While feed-trails like ```malwarepatrol.net``` and ```tallosintelligence.com``` give to this resource softy name ```potential malware site```, let us have precise static trail.


MikhailKasimov:
@stamparm give a hint, pls, why request for ```http://23.249.161.109/``` (e.g. ```curl -I http://23.249.161.109```) doesn't return info on static detection of ```quasarrat.txt```, just keeping on feed-trails info to give?

```Maltrail (server) #v0.10.373``` is in use. Thanks!

maltrail_ 2018-08-30 _-_2018-08-30_16 02 39

PR #265, opened 2018-08-30
Update ipinfo.txt

MikhailKasimov:
The same to ```2ip.ru``` ipinfo site.


stamparm:
I don't see an option from this site to get the user's IP. All those sites inside `ipinfo.txt` should provide back the callers IP (malware workflow)

MikhailKasimov:
Never met malware, which used mentioned resources. OK, let's close for now.

PR #275, opened 2018-08-31
Update phishing.txt

MikhailKasimov:
Addresses from [0] https://twitter.com/PhishingAi/status/1035512536198610949

### DIFF
--- a/trails/static/suspicious/phishing.txt
+++ b/trails/static/suspicious/phishing.txt
@@ -196,3 +196,9 @@ de-agb-token23jei84dh8.com
 # Reference: https://twitter.com/PhishingAi/status/1035509357633359873
 
 att-network.us
+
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
@stamparm Have you an idea how to proceed this entire list from https://www.virustotal.com/#/ip-address/173.249.3.216 ```Passive DNS Replication``` or, if via ```https://cyber-hub.net/passive_ip.php``` web-site (in attach)?

[173.249.3.216_passive_dns.txt](https://github.com/stamparm/maltrail/files/2340215/173.249.3.216_passive_dns.txt)

stamparm:
I have retrieved 1020 alive domains pointing to that same IP. As that one is a shared hosting, I am against blacklisting either that IP or blacklisting all those 1020 domains

PR #284, opened 2018-09-01
Update androm.txt

MikhailKasimov:
[0] https://blog.talosintelligence.com/2018/08/threat-roundup-0817-0824.html ``` Win.Dropper.Delf-6652911-0 ```

Most AVs from VT return ```Androm``` name, starting from ```Kaspersky```.


MikhailKasimov:
Fixed this.

stamparm:
Just googled for `ajmanz.gq` and I can see that lokibot's `/fre.php` is appearing at that same domain. Would not mark this as `androm`

MikhailKasimov:
moved ```ajmanz.gq``` to lokibot static trail.

PR #298, opened 2018-09-01
Create pony.txt

MikhailKasimov:
[0] https://viriback.com/30-days-later-97-panels/

C2 addresses of Pony.


stamparm:
These are admin panels, not C&C addresses. I am against including those as have little to no value compared to C&C addresses

MikhailKasimov:
o_O

MikhailKasimov:
@stamparm After having the second and third look -- how can we exactly know when panel address is equal to C2 address. The logic of panel (some gui-interface) is to manage bots, get statistics from them, put commands for bots and so on... Personally, it is good to detect. Let's recover this trail. M?

PR #312, opened 2018-09-05
Update domain.txt

MikhailKasimov:
Based on #311 entire ```.cl``` domain is added as suspicious.


stamparm:
https://en.wikipedia.org/wiki/.cl:

```
.cl is the Internet country code top-level domain (ccTLD) for Chile
```

PR #318, opened 2018-09-06
Create elf_gafgyt.txt

MikhailKasimov:
```/jackmy*``` was actual both for ```Linux.Kaiten (Tsunami)``` and ```Linux.Gafgyt``` variations. Other names were taken from public resources and my personal monitors. If I can find other names, I\ll update the trail.

@stamparm  Also, please, take a look on [0] https://urlhaus.abuse.ch/host/80.211.189.104/ . Here are also given the most-premost popular filenames for ```Gafgyt``` family:

```/apache2```
```/bash```
```/cron```
```/ftp```
```/ntpd```
```/wget```
```/tftp```
```/openssh```
```/pftp```
```/telnetd```
```/sh```
```/sshd```

but I'm not entirely sure how to detect them without f/a-s.


stamparm:
1) I am against adding regular expression trails into the detection engine as it will introduce problems with performance
2) Looking into the merge proposition I can see problematic trails. Anything that doesn't look at least suspicious and/or really related to the malware should not be included. For example, `/b1`, `/b2`,..., `/i586`, `/i686`, ..., `/xxb1`
3) Along with 2), there are too generic names like `/bins.sh` which doesn't seem like a solid case to connect to the malware (as `gafgyt` in this case)

---

Now, after all being said, I am not sure what to include from here

stamparm:
I'll merge, remove "problematic" trails and commit back with a reference to this pull request

MikhailKasimov:
Thanks!

PR #330, opened 2018-09-08
Update kovter.txt

MikhailKasimov:
Addresses from [0] https://blog.talosintelligence.com/2018/09/threat-roundup-0831-0907.html (Win.Dropper.Kovter-6669952-0 section)


stamparm:
I'll merge these, but majority are down.

Also, after a while (e.g. a month) I'll start filtering out IP address malware trails as they are fastly becoming unusable (for detection purposes)

PR #348, opened 2018-09-11
Update magentocore.txt

MikhailKasimov:
Another ```magecart```: [0] https://www.riskiq.com/blog/labs/magecart-british-airways-breach/

```We saw proof of this on the domain name baways.com as well as the drop server path. The domain was hosted on 89.47.162.248 which is located in Romania and is, in fact, part of a VPS provider named Time4VPS based in Lithuania. ```


stamparm:
Will merge, but those IPs (HTTP) are dead

PR #351, opened 2018-09-12
Update domain.txt

MikhailKasimov:
Based on [0] https://github.com/stamparm/maltrail/pull/350/commits/39b6053ea79f6170249716a5f7b2227712b17f0a VT returns many domains, which are classified as:

```
BitDefender: porn
Dr.Web adult: content
Forcepoint ThreatSeeker: compromised websites
```

So, let entire ```dynu.net``` domain be detected as ```suspicious```.

### DIFF
--- a/trails/static/suspicious/domain.txt
+++ b/trails/static/suspicious/domain.txt
@@ -4948,3 +4948,7 @@ zik.dj
 # Reference: https://twitter.com/ps66uk/status/1037866649435729921
 
 .fun
+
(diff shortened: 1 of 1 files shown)


stamparm:
`dynu.net` is one of main domains used by `dynu.com` dynamic dns provider (https://github.com/stamparm/maltrail/blob/master/trails/static/suspicious/dynamic_domain.txt#L259)

PR #359, opened 2018-09-14
Create sanny.txt

MikhailKasimov:
[0] https://www.fireeye.com/blog/threat-research/2018/03/sanny-malware-delivery-method-updated-in-recently-observed-attacks.html

```SANNY malware uses the FTP protocol as the C2 communication channel```


stamparm:
p.s. removed `ftp://`. Currently Maltrail can accept `IP:port` as trail, but not `domain:port`. `domain:port` is currently impossible as it would require correlating of domain resolution together with connection attempts - currently not possible (primary because packets are processed in different processes - 1 per CPU core)

MikhailKasimov:
OK

PR #372, opened 2018-09-15
Update crypto_mining.txt

MikhailKasimov:
Let is also be here.


stamparm:
`coincola.com` -> mongolock
` drupalupdates.tk ` -> cryptojacking
etc.

crypto_mining should be more or less valid pools for crypto mining. These you wrote look more like malware specific

MikhailKasimov:
OK, will do dedicated malware trails for them.

PR #381, opened 2018-09-15
Update apt_oceanlotus.txt

MikhailKasimov:
[0] https://www.cybereason.com/blog/operation-cobalt-kitty-apt

See screenshot ```Example 2: Cobalt Strike Beacon embedded in obfuscated PowerShell```

### DIFF
--- a/trails/static/malware/apt_oceanlotus.txt
+++ b/trails/static/malware/apt_oceanlotus.txt
@@ -118,3 +118,27 @@ virginiaar.com
 # Reference: https://www.cybereason.com/blog/operation-cobalt-kitty-apt
 
 food.letsmiles.org
+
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
I'm even entirely sure to put this as separate commit. Reason: mentioned in [0] ```teriava.com``` address is already detected by ```apt_oceanlotus``` MT trail. So, I decided not to generate issues. Feel free to move to as separated ```apt_cobaltkitty.txt```, if it is needed.

stamparm:
https://www.cybereason.com/blog/operation-cobalt-kitty-apt <- "cobalt kitty" is just the name of operation. Ocean Lotus is the name of the APT group

PR #384, opened 2018-09-15
Create mongolock.txt

MikhailKasimov:
[0] https://www.bleepingcomputer.com/news/security/mongo-lock-attack-ransoming-deleted-mongodb-databases/


stamparm:
After a second look, `coincola.com` is just an exchange where victim "should" buy the bitcoins to pay the ransom

MikhailKasimov:
So, possibly, to ```crypto-mining.txt``` after all? ;)

stamparm:
`crypto-mining.txt` should also be used to detect in network crypto miners. Sporadic access to sites where crypto mining could be used for ad revenue or access to compromised sites should not be (IMO) captured by Maltrail

PR #385, opened 2018-09-15
Create cryptojack.txt

MikhailKasimov:
[0] https://twitter.com/bad_packets/status/1038705757552660480

### DIFF
--- a/trails/static/malware/cryptojack.txt
+++ b/trails/static/malware/cryptojack.txt
@@ -21,3 +21,7 @@ reauthenticator.com
 # Reference: https://www.virustotal.com/#/url/84743e0afc9055897e1a4aed3cc72058aa345b06452934fd89bcb02e41a276b2/detection
 
 verifier.live
+
+# Reference: https://www.virustotal.com/#/url/e93baa29661c2bcbcd20c9f85eca6f11756d13d829f8703e2efd694635660e7f/detection
+
+gitgrub.pro
--- a/trails/static/malware/cryptojack.txt
+++ b/trails/static/malware/cryptojack.txt
@@ -4,3 +4,8 @@
 # Reference: https://twitter.com/bad_packets/status/1038705757552660480
 
 drupalupdates.tk
+
+# Reference: https://twitter.com/bad_packets/status/1038967603048243200
+
+feesocrald.com
+soodatmish.com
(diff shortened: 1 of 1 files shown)


stamparm:
After a closer look I can see that this could go way beyond the "malware" story. I am against adding of mining JS scripts on compromised sites as those are not a good reference for "malware infection".

MikhailKasimov:
50/50 here. This can give a hint to admin, that site compromised is and has a connection to some crypto-mining staff. Can be useful.

stamparm:
https://docs.google.com/spreadsheets/d/1RdT_r4fi4wPx5rY306FftVKaXiAZeQeb5fx78DmbVx0/edit#gid=0

Based on that table somebody could blacklist (e.g.) Coinhive and say that it is a malicious service. I am not happy with that idea

MikhailKasimov:
Here is again 50\50, just as a case with ```PUP``` : when someone uses crypto-mining on its own pc/laptop -- it's OK. When via someone else's ones -- it is evil and should be detected as a marker, that something is going wrong. Personally I'm OK to be alerted.

stamparm:
`crypto_mining` should be to detect crypt mining voluntary (e.g. breaking of company's policy) or involuntary (e.g. malware).

Now, sporadic running of crypto JS (most probably involuntary) by visiting some compromised site or site which wants to generate some extra non-Ads revenue, will generate noise IMO.

MikhailKasimov:
```(most probably involuntary)``` <-- that's the key phrase, IMHO. Personally, I have already two compromised sites to clean exactly from ```JS```-miners, which slow down sites, but hoster's tools are silent. And I'm looking for way to have simple beacon, which will give me a hint, that site has begun connections related to crypto-mining. And to block such connections on firewall. That's all. Noise? Possibly, but let us have some live tests to re-check this out and rate of such noise. M?

MikhailKasimov:
BTW, addresses I put in PR here are not addresses of compromised sites, but sites, which these compromised sites try to connect to.

MikhailKasimov:
https://twitter.com/bad_packets/status/1042630612853514240

MikhailKasimov:
Resolved via ```malicious_js``` mechanism  ( #459 ).

```/lib/crypta.js``` , ```/bitrix/js/main/core/core_loader.js``` and ```/bitrix/js/main/core/core_tasker.js``` are here: https://github.com/stamparm/maltrail/pull/460/commits/67e0071f11e45ce7b53fd2412f01f25c05a8d3c2

PR #405, opened 2018-09-20
Update phishing.txt

MikhailKasimov:
Some updates for 2018-09-20 from Twitter accounts.

### DIFF
--- a/trails/static/suspicious/phishing.txt
+++ b/trails/static/suspicious/phishing.txt
@@ -616,3 +616,31 @@ add-me-sc.com
 # Reference: https://twitter.com/nullcookies/status/1042537633958658048
 
 dope.org/9x/9x/
+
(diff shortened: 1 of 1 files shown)


stamparm:
Majority of those after a last reference look like compromised sites

p.s. unrelated. Please triple check any threats coming from those "threat-roundups" coming from `https://blog.talosintelligence.com/` as I have more and more feeling that they are doing a POOR/AMATEUR job. It was not a first time to put REGULAR TOP domains into their "threat" analyses (e.g. https://github.com/stamparm/maltrail/commit/e6d8be382c8e903c7f30e780e813274a67681f67)

MikhailKasimov:
Hmm... OK, will take a more closer look on info from this source. Thanks for a hint!

PR #457, opened 2018-09-24
Create wordpress_generic.txt

MikhailKasimov:
[0] https://blog.malwarebytes.com/threat-analysis/2018/09/mass-wordpress-compromises-tech-support-scams/

Attack on WP has no name, so ```wordpress_generic.txt```.

PR #459, opened 2018-09-24
Create wordpress_compromised.txt

MikhailKasimov:
[0] https://github.com/stamparm/maltrail/pull/457#issuecomment-423946039

### DIFF
--- a/trails/static/suspicious/wordpress_compromised.txt
+++ b/trails/static/suspicious/wordpress_compromised.txt
@@ -1,0 +1,11 @@
+# Copyright (c) 2014-2018 Miroslav Stampar (@stamparm)
+# See the file 'LICENSE' for copying permission
+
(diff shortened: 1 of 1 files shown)


stamparm:
@MikhailKasimov now, here comes the second thought :)

As user will just see the final loading of the "malicious" JS (as trails are in such cases e.g. hosts containing such JS files), please do the `suspicious/malicious_js.txt`

I can do that (after a merge) if that's ok with you?

MikhailKasimov:
@stamparm

``` As user will just see the final loading of the "malicious" JS``` --> it's a really good chance to have #385 also being merged, because logic is similar here and there. ;)

  ```I can do that (after a merge) if that's ok with you?``` --> personally, everything is OK, especially when it begins to detect malicious trails. :)

MikhailKasimov:
UPD: Well, ```suspicious/malicious_js.txt``` trail can even have ```/lib/crypta.js``` (see also #385 ) callback as the trail.

PR #468, opened 2018-09-25
Update crypto_mining.txt

MikhailKasimov:
[0] https://twitter.com/bad_packets/status/1038967603048243200


stamparm:
Please put this into the `malicious_js.txt`. Web browser mining should go there. `crypto_mining.txt` should be used ONLY for mining pools. We discussed this already - cryptojacking is malicious JS

PR #476, opened 2018-09-27
Update apt_c23.txt

MikhailKasimov:
All deleted dup domains are mentioned above as I can see.


stamparm:
There is no need for doing this at the single `trail.txt` basis as Maltrail automatically uses unique items from each file. It is slightly better practice to have single reference with IoCs afterwards than to have one big pile of domains with lots of different references. Sometimes this is not possible, but as said, there is no need for merging stuff inside a single trail file

MikhailKasimov:
ok

PR #481, opened 2018-09-27
Create jscoffe.txt

MikhailKasimov:
[0] https://www.symantec.com/security-center/writeup/2018-092007-1208-99

It is not ```suspicious```, but ```malware``` instead case.

PR #482, opened 2018-09-27
Update malicious_js.txt

MikhailKasimov:
One domain from  #481 has the open dir -- ```energycoffe.org/js/```. Feel free to move ```/js/*``` to #481 if needed.

UPD: ```coffetea.org/js/``` has the same pack of js.

UPD2: After the second look, I'm standing for moving that to #481 because it is uniquely named malware. M?


MikhailKasimov:
Possibly, merge conflict is the hint to move related to ```jscoffe``` js to ```jscoffe.txt``` trail.

Issue #313, opened 2018-09-05
Idea for 'long .php (suspicious)' heur trail


MikhailKasimov:
Hello!

Looking on long random-sequenced name ```/tnmvqxxf9bmua2a6qbhpvc7k.php``` (see [0] https://twitter.com/PhishingAi/status/1037158782415994880) and ```//PvqDq929BSx_A_D_M1n_a.php``` (see ```lokibot.txt``` trail), I have an idea-question: can be this snippet of code from ```sensor.py``` for long domains heur detection also be applicable (after its modification, of course) to detect anomalies in php-endings like ```/something-long-and-random.php``` ?

```
if not result and config.USE_HEURISTICS:
            if len(parts[0]) > SUSPICIOUS_DOMAIN_LENGTH_THRESHOLD and '-' not in parts[0]:
                trail = None

                if len(parts) > 2:
                    trail = "(%s).%s" % ('.'.join(parts[:-2]), '.'.join(parts[-2:]))
                elif len(parts) == 2:
                    trail = "(%s).%s" % (parts[0], parts[1])
                else:
                    trail = query

                if trail and not any(_ in trail for _ in WHITELIST_LONG_DOMAIN_NAME_KEYWORDS):
                    result = True
log_event((sec, usec, src_ip, src_port, dst_ip, dst_port, proto, TRAIL.DNS, trail, "long domain (suspicious)",

```

stamparm:
I have a feeling that this could cause lots of false-positives. I can check tomorrow at live environment

MikhailKasimov:
OK, thanks!

stamparm:
One day traffic (data for only unique PHP files - Note: including extension `.php`):

```
# NumSamples = 1904; Min = 5.00; Max = 61.00
# Mean = 13.554097; Variance = 38.075855; SD = 6.170564; Median 12.000000
# each ∎ represents a count of 11
    5.0000 -    10.6000 [   645]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
[...]

PR #488, opened 2018-09-28
Create malicious_php.txt

MikhailKasimov:
[0] https://www.virustotal.com/#/ip-address/85.17.26.65 (#URL section)

The idea is similar to ```malicious_js.txt``` trail and is a some rethought of #313 -- to detect suspicious ```.php```.

### DIFF
--- a/trails/static/suspicious/malicious_php.txt
+++ b/trails/static/suspicious/malicious_php.txt
@@ -24,3 +24,7 @@
 /psycho.php
 /review_me.php
 /rootme.php
+
+# Reference: https://www.virustotal.com/#/domain/manapowermta.us
+
+/loomistech/gate.php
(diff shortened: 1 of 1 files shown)


stamparm:
I believe that you are over-reaching now. It is great that you want to cover lots of stuff, but there is no need to cover everything. You'll burn out.

Malicious PHP can have any arbitrary name. As I've seen, if not strictly malware related (e.g. like `/gate.php`), bad guys choose really fascinating names. If not malware related (i.e. they can't be related to some malware strain) I would not do this kind of stuff

MikhailKasimov:
Well, I took that ones, that are related to malware or phishing only. Otherwise, I would put ```index.php```, ```login.php``` and so on. :) And in this case yes, that will cover huge pile of clean resources. That's why I try to put other parts of path (e.g.: ```/1/``` , ```/r/``` and other sign combinations), to make detection more specific instead of wide.

MikhailKasimov:
So?

MikhailKasimov:
@stamparm Thanks!

PR #489, opened 2018-09-28
Update phishing.txt

MikhailKasimov:
[0] https://www.virustotal.com/#/ip-address/85.17.26.65

```sort```-ed and ```uniq```-ed phishing domain-list.

### DIFF
--- a/trails/static/suspicious/phishing.txt
+++ b/trails/static/suspicious/phishing.txt
@@ -985,7 +985,6 @@ takipcinsta.aba.ae
 takipfollow.com
 unicbl67410.aba.ae
 upload-information.aba.ae
-yopi-car.com
 zero.zz.com.ve
 
 # Reference: https://twitter.com/smica83/status/1045587050932695040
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
Removed FP for yopi-car.com

stamparm:
I said at least. Lots of domains from passive DNS lookup look like false-positives. Closing this down

MikhailKasimov:
``` look like``` and other ```highly likely```-s... And closing with other real phishing domains like ```bcft.aero/ID/PayPal/Help/``` and later. You bet.

MikhailKasimov:
E.g.:

``` facebox.aba.ae
faxebok.aba.ae```

look like false-positives, aha...

MikhailKasimov:
I've rechecked these dimains on VT, I have doubts on ```zero.zz.com.ve``` and ```pdfdoc.aba.ae``` only. If you can see other clean stuff, please, give a hint. Thanks!

PR #498, opened 2018-09-29
Update njrat.txt

MikhailKasimov:
[0] https://twitter.com/ScumBots/status/1045701425198698499


stamparm:
This IP is currently down. Port can't be communicated with. RATs like usage of (dynamic) domains. Putting some potential DSL IP in here doesn't make any sense

PR #508, opened 2018-09-30
Update lokibot.txt

MikhailKasimov:
[0] https://twitter.com/ViriBack/status/1046391838448537601
[1] https://pastebin.com/4QRaU8T7

We simultaneously accept ```Loki``` and decline ```Pony``` trails, like it was in #298 . What should we do for ```Pony``` after all?

### DIFF
--- a/trails/static/malware/lokibot.txt
+++ b/trails/static/malware/lokibot.txt
@@ -412,6 +412,63 @@ grace4good.cf
 theonlygoodman.com
 premierevents.co.zw
 
+# Reference: https://twitter.com/ViriBack/status/1046391838448537601
+# Reference: https://pastebin.com/4QRaU8T7
+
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
@stamparm BTW, need your decision on this: https://pastebin.com/uRkNCYYn

stamparm:
Pony:

1) http://cybercrime-tracker.net/index.php?search=Pony <- will give you ADMIN panels
(e.g. take tolain.ru which is quite popular these days)
2) https://www.virustotal.com/#/file/24497a76af9ecbf66d915631ed766ced2c7d6cb78720f68e91fdf234506fa7dc/behavior will show you that the related file (i.e. malware) is communicating with the `GATE` of that same domain and/or compromised site

There is no point in putting ADMIN panel URLs into the trails. PERIOD.

PR #525, opened 2018-10-01
Update zeus.txt

MikhailKasimov:
[0] https://pastebin.com/bPV4gVVL (pandabanker)

PR #536, opened 2018-10-02
Update pandabanker.txt

MikhailKasimov:
@stamparm You've forgotten to move here these trails from #525 you closed. Double work again.


stamparm:
Nope. I created an independent collection of trails from relevant sources

PR #545, opened 2018-10-02
Update trickbot.txt

MikhailKasimov:
[0] https://myonlinesecurity.co.uk/trickbot-still-being-delivered-by-fake-payroll-emails/


stamparm:
Those two sites look like compromised

MikhailKasimov:
First - yes, agreed. The second returns 403, but, to avoid f/p, is currently deleted. Generic trail is actual.

PR #609, opened 2018-10-05
Update apt_sofacy.txt

MikhailKasimov:
United trail for dedicated ```xtunnel.txt```, ```computrace.txt```, ```chopstickrat.txt``` trails, which were created to detect specified tools.


stamparm:
`computrace.txt` is not just APT28 based. In my place there would be tons of FP. Also,

PR #607, opened 2018-10-05
Create computrace.txt

MikhailKasimov:
[0] https://twitter.com/Bank_Security/status/1048113406597910528
[1] https://www.ncsc.gov.uk/content/files/protected_files/article_files/IOC-APT28-malware-advisory.pdf (CompuTrace chapter)

Domain, met in Snort rule, only. Feel free to add IP-list, if it is acceptable.

PR #614, opened 2018-10-05
Update computrace.txt

MikhailKasimov:
Moving domain from https://github.com/stamparm/maltrail/pull/607/commits/726c4f159cd0dcef5f67b7a476e6d8baf668d415


stamparm:
`fnbcorporate.co.za` doesn't have anything to do with Computrace. It is just part of that Snort rule NCSC gave to exclude false positives (most probably related HTTP traffic to that same domain contains `TagId` inside the content)

PR #626, opened 2018-10-07
Update whitelist.txt

MikhailKasimov:
Cisco's FPs on https://blog.talosintelligence.com/2018/10/threat-roundup-0928-1005.html (Win.Malware.Gandcrab-6706045-0)

### DIFF
--- a/misc/whitelist.txt
+++ b/misc/whitelist.txt
@@ -38,18 +38,6 @@ cdxall.com
 cdxgoog.com
 cdxultra.com
 cedexis-radar.net
-
-# Cisco's FPs on https://blog.talosintelligence.com/2018/10/threat-roundup-0928-1005.html (Win.Malware.Gandcrab-6706045-0)
-
-asl-company.ru
-bethel.com.ve
-dreamhost.com
-help.dreamhost.com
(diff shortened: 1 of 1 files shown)


stamparm:
There is no need for updating whitelist based on FP at some reference. Just not inclusion of FP into trails is enough. If you insist I'll do it, but in future, as said, there is no need for doing this

MikhailKasimov:
Anyway, it is desirable to add these ones.

```but in future, as said, there is no need for doing this``` -- agreed. Thanks!

MikhailKasimov:
M?

stamparm:
Because of #627, dropping this down. If you insist, you can re-include some of those to the whitelist (as you did previously), but please drop the talosintelligence reference.

MikhailKasimov:
```but please drop the talosintelligence reference.```

Looking into ```whitelist.txt```:

```# appeared on malwarepatrol.net```
```# appeared on malwaredomainlist.com```

So, how should I name such PR, if ``````but please drop the talosintelligence reference.```

P.S. On ```but please drop the talosintelligence reference.``` -- no problem. As I said above -- agreed.

PR #635, opened 2018-10-08
Update brushaloader.txt

MikhailKasimov:
[0] https://twitter.com/Racco42/status/1049233960998514689


stamparm:
Looks too generic. Just by taking a look into those trails (without knowing that it might be something related to the `brushaloader`) I would never say that it is malware related

MikhailKasimov:
```N``` and ```M``` in trails parameters are random. Not aware how this can be detected reliably with MT without regexp support.

PR #638, opened 2018-10-08
Update elf_mirai.txt

MikhailKasimov:
The same IP to ```/juno``` trail (See [0] http://39.104.190.48/D%3A/p-an/Tomcat7/logs/localhost_access_log.2018-10-02.txt).

### DIFF
--- a/trails/static/malware/elf_mirai.txt
+++ b/trails/static/malware/elf_mirai.txt
@@ -215,3 +215,4 @@ ukrainianhorseriding.com
 # Generic trail from MT heuristic detection
 
 /juno
+/cock


stamparm:
Can't associate some random dictionary word (`cock` in this case) with Mirai. Just look at the trail and try to deduce that it is "Mirai"

MikhailKasimov:
```Can't associate some random dictionary word``` --> so, you have even taken a look into [0]? It is not ```random dictionary word```, it is part of link to download.

```Just look at the trail and try to deduce that it is "Mirai"``` --> without external link on log (like [0]), yes. But not in this case. So, I would insist on detection.

@stamparm

PR #641, opened 2018-10-09
Update malicious_php.txt

MikhailKasimov:
[0] https://twitter.com/bad_packets/status/1049511435384967168
[1] https://pastebin.com/raw/Arzk2D76

```sort```-ed and ```uniq```-ed list. Also some too generic trails, like ```index.php```, ```log.php```, ```test.php``` and so on, were deleted.

If some of trails are also ```too generic```, let them cut off too.


stamparm:
Majority of those are too generic. Sorry, can't include this. Also, one important note. Random made up names used for random dropping of malicious PHP scripts does not mean that it will be used AGAIN by that same made up name at some other (e.g. compromised) site

MikhailKasimov:
@stamparm  ```Majority of those are too generic. Sorry, can't include this.```--> Miroslav, I have a very intelligent idea on such cases: let us not close entire PR as well, but filter that ones, which can be applicable. I understand, that ```Random made up names used for random dropping of malicious PHP scripts does not mean that it will be used AGAIN by that same made up name at some other (e.g. compromised) site```, but, e.g. ```/ololo123.php``` can be met on hundreds (compromised) sites. And second iteration ```/shmololo321.php``` will, certainly, break our detection, but **I strongly convinced*** to detect as ```suspicious```  random ```/ololo.123.php``` even if it won't be met any more.

So, let's take some time to filter things, which are not ```too generic``` here, and add them. Thanks!

PR #657, opened 2018-10-11
Update brushaloader.txt

MikhailKasimov:
[0] https://twitter.com/pollo290987/status/1050384298048704514


stamparm:
Too generic. Also, by just Googling around found https://github.com/understrap/understrap/blob/master/page-templates/blank.php

MikhailKasimov:
Let's invent some method ti detect ```brushaloader```

PR #665, opened 2018-10-12
Create darkcoders.txt

MikhailKasimov:
[0] https://twitter.com/ViriBack/status/1050543003935416321

### DIFF
--- a/trails/static/malware/darkcoders.txt
+++ b/trails/static/malware/darkcoders.txt
@@ -1,0 +1,7 @@
+
+# Copyright (c) 2014-2018 Miroslav Stampar (@stamparm)
+# See the file 'LICENSE' for copying permission
+
+# Reference: https://twitter.com/ViriBack/status/1050543003935416321
+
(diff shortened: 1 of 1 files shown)


stamparm:
Just a login page. Discussed this kind of cases multiple times. C&C callbacks have value, login pages don't

MikhailKasimov:
50/50. If someone from internal network connects to such login page, this can be considered as a semaphore of security incident. Personally, I'll appreciate such functionality.

stamparm:
Possibility of a someone to get infected: very slim. Possibility of a someone to actually be a botnet operator (inside your network): zero to nil. Already discussed before

MikhailKasimov:
Again, here is 50/50: it can be a operator from IT dep or IS dep. Nil -- possibly, but I wanna to have such detection although as ```suspicious``` one, even in case of being ```zero to nil```.

@stamparm  We can have a generic trail in ```suspicious``` folder, named, e.g., ```malicious_panel.txt```. And I'm insisting on it here.

PR #681, opened 2018-10-13
Update crypto_mining.txt

MikhailKasimov:
[0] https://www.cyren.com/blog/articles/facebook-users-threatened-with-spam-run-delivering-crypto-miner-malware

Really don't know what to choose here: ```malicious_js.txt``` <-> ```malicious_php.txt```, but I also do not want to split trails. If ```delivering-crypto-miner-malware```, then let it be put to ```crypto_mining.txt```. If you have doubts on it -- ok, let's discuss.

### DIFF
--- a/trails/static/suspicious/crypto_mining.txt
+++ b/trails/static/suspicious/crypto_mining.txt
@@ -90,3 +90,13 @@ pool.cortins.tk
 pool.supportxmr.com
 xmr.crypto-pool.fr
 xmrpool.eu
+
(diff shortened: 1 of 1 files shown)


stamparm:
The whole article repeats "threat actors". Also, given trails/IoCs are mostly steps in infection chain. I would either use (new name) "KromAgent" which they refer to in couple of occasions or just put everything in generic.

I prefer `kromagent.txt` in this moment

MikhailKasimov:
Superseded by #709

PR #712, opened 2018-10-17
Update generic.txt

MikhailKasimov:
[0] https://twitter.com/nullcookies/status/1052355899422052352 (root oldies.esy.es is going to generic.txt)


stamparm:
`generic.txt` is for malware, while this site/domain looks like phishing-only

MikhailKasimov:
@stamparm Take a look on https://github.com/stamparm/maltrail/pull/713/commits/761fcfedd112c0741704974b1cb6afe9f9340b08 , where the first root one is compromised site and the second one is malicious instead.

PR #717, opened 2018-10-17
Update lokibot.txt

MikhailKasimov:
[0] https://twitter.com/avman1995/status/1052426452187185153 + generic trail update.


MikhailKasimov:
Do not be so scared. :) That's why I've put both: domain-based and generic-based trails: to have a main and additional trails. And I really have a tons of doubts, that ```/chri1/cgi.php``` ca be met on legal sites. ```/cgi.php``` -- yes, but not ```/chri1/cgi.php``` one.

PR #752, opened 2018-10-20
Update elf_generic.txt

MikhailKasimov:
[0] https://twitter.com/stefant/status/1053227806258327558

PR #753, opened 2018-10-20
Update elf_coinminer.txt

MikhailKasimov:
Moved from [0] https://github.com/stamparm/maltrail/pull/752/commits/20e635539e30476a29d910a0ea715f1145c58490


stamparm:
If I would see those names unrelated to the `elf_coinminer` I would never say that those are "malware" related

PR #761, opened 2018-10-20
Update elf_billgates.txt

MikhailKasimov:
Based on samples, found here: ```123.249.71.250:666``` (https://github.com/stamparm/maltrail/pull/760/files#diff-f8c0b14542878c134c3bd3925df83f52). If need, I can give a screenshot. Especially for such persons, who have a fruity habit to close PRs with lousy\fragile arguments, like ```If I would see those names unrelated to the [some-trail] I would never say that those are "malware" related``` (c)

### DIFF
--- a/trails/static/malware/elf_billgates.txt
+++ b/trails/static/malware/elf_billgates.txt
@@ -13,3 +13,11 @@ liunx.200jh.com
 # Reference: https://otx.alienvault.com/pulse/557b7ca6b45ff5450f63f4f5/
 
 wuzu520.com
+
+# Generic trails
+
+/ddos2.4
+/ddos32-64
+/Linux2.6
+/syn25000
+/udp25000


MikhailKasimov:
elf_gates

stamparm:
I'll close this down because of your language. I've tried to discuss hundreds of times that something lke:

`Linux2.6` is too generic to be associated with for example `elf_billgates`. Malware authors and/or operators use RANDOM/GENERIC names out of their head. That doesn't mean that everything found at single place can be used as "trail".

If I would accept all those trashy names, maltrail would simply become a trash depo. Closing this down and also stopping future merge requests. Not sure why do you need to fight with me???

MikhailKasimov:
@stamparm 

```Not sure why do you need to fight with me???```

Not with you personally, just with your broken double-twisted logic of detection (hope, you remember hidtory of IP detection and how I caught you?). Especially, when you close entire PR, when it contains one FP among hundreds of trails, instead of telling something like ```delete this and others are good```. And this makes me out of any civil language. So, stop crying on someones ```aggressive behavior``` and let us begin our cooperation in more adequate key from your personal side first of all.

```Now, you are blaming me to close down requests with clearly generic/random/shitty names BECAUSE to not offend you in any case???``` --> I'm not aware on ```generic/random/shitty names```, I am aware of these filenames, which are in use in MALWARE. And I really do filtering before to add some generic trail. And, when I have doubts, I tell about them (see, e.g. ```apt_dustsquad.txt```).

```  BECAUSE to not offend you in any case???``` --> Wrong opinion.

```Please, don't do any more merge requests``` --> Again, stop crying. I could leave putting PRs many times ago, especially after your first closing PRs with lousy arguments. But I still work on making MT good as IDS. And going to do that further.

PR #777, opened 2018-10-22
Update maltrail.conf

MikhailKasimov:
```#DISABLE_CHECK_SUDO true``` is ubuntism, IMO. User/admin should really understand what does he do, when disabling ```sudo``` check. Should be ```false``` by default. It is familiar on manual putting ```yes``` in console on ssh-connections.

### DIFF
--- a/maltrail.conf
+++ b/maltrail.conf
@@ -107,4 +107,4 @@ LOG_DIR $SYSTEM_LOG_DIR/maltrail
 #PROXY_ADDRESS http://192.168.5.101:8118
 
 # Disable checking of sudo/Administrator privileges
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
@stamparm Arguments, discussion?

stamparm:
It is a commented out option. Hence, I don't see a great deal here. If I (user) am un-commenting it, I would expect something to do with default setting. If default (proposed) `DISABLE_CHECK_SUDO false` is used, it won't do any changes in program behavior.

Anyway, commented out option should not be such a great deal to discuss

MikhailKasimov:
OK. In this case let's close it now. Thanks!

PR #792, opened 2018-10-26
SameSite and secure cookie flags

D9ping:
@stamparm This PR adds the SameSite and the Secure cookie flags to the session cookies to limit the damage of cross-site request forgery and improve the security of the browser session.


stamparm:
@D9ping thank you and sorry for delay

PR #795, opened 2018-11-09
Correct the PCAP_FILE parameter

borjam:
It is very useful to allow sensor.py to open a pcap file. Not only for offline analysis, but also allowing the sensor to receive pcap data streamed from another location. In that case /dev/stdin works. 

An example application is packet capture from Mikrotik routers. The builtin packet sniffer includes a streaming option that sends the captured packets over UDP using the TZSP protocol.

[https://en.wikipedia.org/wiki/TZSP](url)

### Example setup:

Just install tzsp2pcap [https://github.com/thefloweringash/tzsp2pcap](url)

on maltrail.conf set PCAP_FILE to /dev/stdin

and run `tzsp2pcap -f | python ./sensor.py`

### Example on the Mikrotik router, rather self explanatory.

`[admin@router] /tool sniffer> export 
/tool sniffer
set filter-interface=ether10.24 filter-operator-between-entries=and \
    streaming-enabled=yes streaming-server=192.168.1.202
[admin@router] /tool sniffer> 
`

### DIFF
--- a/maltrail.conf
+++ b/maltrail.conf
@@ -57,6 +57,13 @@ CUSTOM_TRAILS_DIR ./trails/custom
 # (Max.) size of multiprocessing network capture ring buffer (in bytes or percentage of total physical memory) used by sensor (e.g. 512MB)
 CAPTURE_BUFFER 10%
 
+#
+# PCAP file to monitor instead of an actual network interface.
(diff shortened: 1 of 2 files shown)


stamparm:
@borjam `pcap_file` (in original) is actually provided with option `-i`. Hence, something like this should work in your case: `tzsp2pcap -f | python ./sensor.py -i -`

PR #838, opened 2019-03-04
Update cobalt.txt

MikhailKasimov:


stamparm:
Should be transformed into the feed based on https://github.com/fox-it/cobaltstrike-extraneous-space/blob/master/cobaltstrike-servers.csv. Will do it tomorrow (if I catch time)

stamparm:
1) Now there is a new feed ([cobaltstrike.py](https://github.com/stamparm/maltrail/blob/master/trails/feeds/cobaltstrike.py))
2) Cobalt (Gang) != Cobalt Strike
3) Trails are being took from Fox-It's list if `last_seen` within last 2 months
4) Detection inside the traffic is being marked as `cobalt strike (adversary)` with `HIGH` severity <- it means that somebody (adversary) is using the "Cobalt Strike" against the org's infrastructure

PR #839, opened 2019-03-05
Create apt_sharpshooter.txt

MikhailKasimov:

PR #841, opened 2019-03-05
Update and rename apt_unknown.txt to apt_generic.txt

MikhailKasimov:
Trails, moved from #839 . Also ```apt_unknown.txt``` -> ```apt_generic.txt```. Personally, ```generic``` sounds better, than ```unknown```.


stamparm:
IMO, "unknown APT" actor sounds better (more serious) than "generic APT" actor

p.s. if you wanted to reference the usage in case of "generic malware" <- malware in that case doesn't have a "serious" note as in case of APT actor

MikhailKasimov:
> IMO, "unknown APT" actor sounds better (more serious) than "generic APT" actor
> p.s. if you wanted to reference the usage in case of "generic malware" <- malware in that case doesn't have a "serious" note as in case of APT actor

OK, have reverted to ```unknown```.

PR #845, opened 2019-03-05
Create bitter.txt

MikhailKasimov:


stamparm:
https://github.com/stamparm/maltrail/blob/master/trails/static/malware/apt_bitter.txt

MikhailKasimov:
```Ctrl-F``` didn't return, that ```bitter``` is also present, sorry. :( Will move generic trails to ```apt_bitter.txt```

PR #859, opened 2019-03-07
Create carrotbat.txt

MikhailKasimov:


stamparm:
-> `apt_unknown.txt`

MikhailKasimov:
Hmmm... even if malware name is given explicitly in initial article?

stamparm:
We had a similar discussion couple of days ago [here](https://github.com/stamparm/maltrail/pull/839#issuecomment-469626741).

Even `carrotbat` is better than `apt_carrotbat` if you insist on having a name. I just wanted to mark it as `apt_` as it has an `APT` actor behind it.

MikhailKasimov:
#861

PR #863, opened 2019-03-08
Create thbaiai_banker.txt

MikhailKasimov:
C2 addresses. Domain list is in #862

### DIFF
--- a/trails/static/malware/thbaiai_banker.txt
+++ b/trails/static/malware/thbaiai_banker.txt
@@ -1,0 +1,7 @@
+# Copyright (c) 2014-2019 Maltrail developers (https://github.com/stamparm/maltrail/)
+# See the file 'LICENSE' for copying permission
+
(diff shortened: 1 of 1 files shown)


stamparm:
Why #862 and #863 in different files?

stamparm:
p.s. I would just leave `generic.txt` as that same domain is/will die (hence, no HTTP requests toward it)

MikhailKasimov:
> 
> 
> Why #862 and #863 in different files?

Because domain ```lt99.ddns.net``` and  ```/al/index.php```/```logsD/index.php``` trails, when separate, seem to be "too generic" (c). Hence, I decided to put full trail paths as the explicitly defined trail for this banker-trojan.

stamparm:
Domain is dead (as said before), hence there will be no HTTP requests. `generic` has been already merged

PR #867, opened 2019-03-09
Update njrat.txt

MikhailKasimov:


stamparm:
Going to merge, but as said in previous comments, [blackbook](https://github.com/stamparm/blackbook/blob/master/blackbook.csv) already covers majority of ScumBots trails

MikhailKasimov:
@stamparm As I can see inside ```blackbook```, ```IP:Port``` for ```njrat``` anyway should be added manually, OK, will check them periodically and ```blackbook``` will check ```njrat``` domains in ScumBots.

PR #876, opened 2019-03-09
Update xtrat.txt

MikhailKasimov:


stamparm:
Already inside [blackbook](https://github.com/stamparm/blackbook/blob/master/blackbook.csv) - used feed

PR #892, opened 2019-03-11
Create malicious_pos_forums.txt

MikhailKasimov:
The idea of this trail is to detect as ```suspicious``` connections to various point-of-sale forums, where malicious programs can be sold/exchanged, etc. Not sure if there is any public trail for such purposes, but let's have such one: this can be useful for security-admins.


stamparm:
-> `malicious_forum.txt`

MikhailKasimov:
@stamparm done.

PR #895, opened 2019-03-12
Update apt_chafer.txt

MikhailKasimov:
See ```Control server HTTP URL``` in ```Configuration``` table.


stamparm:
You know that I am not a fan of IP trails. This one is from 2018, hence, nobody would miss it if not merged

MikhailKasimov:
Just as historical one.

stamparm:
WHOIS for that same IP is `Choopa, LLC` (VPS provider). It most probably changed couple of owners till now. I had problems with this kind of "trails" when CDNs don't care about the reputation of IPs (buying whatever they can for least possible price)

MikhailKasimov:
OK.

PR #916, opened 2019-03-14
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
I am removing `afraid.org` as it is a provider of dynamic domains, but not directly a eTLD for dynamic domains like in other cases. This basically means that by seeing it inside the Maltrail you won't be able to get the malicious domain per-se.

We can further discuss. Until then merging the rest

stamparm:
Based on #919, we could also put `afraid.org`, as I didn't know that they are also providing eTLD for their own domain, but I am not very keen of the idea. Anyway, just `afraid.org` should suffice (without those `ns1.`, `mail.`,...)

MikhailKasimov:
Initially my idea was move all clean ```*.afraid.org``` domains (like ```ns.```, ```mail.``` etc) to ```dynamic_domain``` and other stuff to keep as malicious. Due to #919 I can clean the list up and to add all other as generic ```gtbot```. Please, reopen #919

stamparm:
1) `dynamic_domain` should be used for eTLDs (right most domain part of the dynamic domain name), not for marking "regular" services. For that kind of stuff potentially you could use `misc/whitelist.txt`
2) Domains in #919 are copy-pasted from ` # Reference: https://www.virustotal.com/#/domain/afraid.org`. Please use approach like in https://github.com/stamparm/maltrail/pull/919#issuecomment-473202226 if you want to use VirusTotal as a reference for gtbot

MikhailKasimov:
1. OK

2. Your list is different to ```gtbot.txt - clean trails```. I wanna to unite these two lists.

PR #923, opened 2019-03-15
Update dynamic_domain.txt

MikhailKasimov:

PR #924, opened 2019-03-15
Update dynamic_domain.txt

MikhailKasimov:
Resounding of #923


stamparm:
There no dynamic-alike subdomains under `m-a-f-i-a.com` (Reference: https://www.virustotal.com/#/domain/m-a-f-i-a.com)

PR #919, opened 2019-03-14
Create gtbot.txt

MikhailKasimov:

PR #925, opened 2019-03-15
Create gtbot.txt

MikhailKasimov:
Merged lists from #919 . Also ```afraid.org``` list was updated to detect malicious trails (clean trails were removed). Also ```gtbot``` is currently generic naming and, to check correct name for each trail is redundant. Or malicious ```afraid.org``` trails can be removed to ```generic.txt``` if needed.


MikhailKasimov:
OK, will re-work ```afraid.org``` list again, no problem, leaving trails with ```Communicating files``` in VT (e.g. https://www.virustotal.com/#/domain/1liil11liil1.afraid.org). But I really can get, who can invent ```crunch.*``` name for legitimate purposes. :)

stamparm:
I am using afraid.org service for my own private purposes. I have such names that they all appear "suspicious". As a side note, just usage of dynamic-domain name provider raises a suspicion, but can't be marked immediately as malicious.

MikhailKasimov:
That's why I will leave ones, like https://www.virustotal.com/#/domain/1liil11liil1.afraid.org (where Communicating files is present and such files are excplicitly malicious). Stay tuned. :)

PR #931, opened 2019-03-15
Create sdbot.txt

MikhailKasimov:
Resounded from #919 , attempt number 3.  ```sdbot``` name is met most of the times for these domains due to VT files detection. Each trail contains its ```Communicating Files``` or ```Files Referring``` section on VT respectively.

### DIFF
--- a/trails/static/malware/sdbot.txt
+++ b/trails/static/malware/sdbot.txt
@@ -1,0 +1,41 @@
+# Copyright (c) 2014-2019 Maltrail developers (https://github.com/stamparm/maltrail/)
+# See the file 'LICENSE' for copying permission
+
+# Reference: https://www.virustotal.com/#/domain/*.afraid.org
+
+1liil11liil1.afraid.org
+1llsklll.afraid.org
+b0ts.afraid.org
+bent.afraid.org
+blardy.afraid.org
+demrt.afraid.org
+eleven.afraid.org
+exaturbo.afraid.org
+fjpjllj1lx.afraid.org
+four.afraid.org
+franck78.afraid.org
+fuckthebac0n.afraid.org
+imiill11lnot.afraid.org
+hlph0pfIipf10p.afraid.org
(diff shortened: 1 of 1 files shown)


stamparm:
I can see that they are malware related. Went through couple of them in Virustotal and can't find `BOT` `gbot` nor `sbot` inside.

Please, -> `generic.txt`

MikhailKasimov:
> 
> 
> I can see that they are malware related. Though, went through couple of those at Virustotal and can't find `BOT`, `gtbot` nor `sdbot` inside.
> 
> Please, -> `generic.txt`

One more chance before moving to ```generic.txt```: Take https://www.virustotal.com/#/domain/1liil11liil1.afraid.org , for example, and even the first file from ```Communicating Files``` --  https://www.virustotal.com/#/file/bc5662f91493da2bd02781832f8728a3eba350e47ed11cd8f927e3cb88a44452/detection : you can meet and ```GTBot```, and ```SDBot```, and ```Aebot``` in vendor's detections.

stamparm:
1) If there is already a `generic.txt`, I don't see a reason to use some other "generic" naming (e.g. `SDBot` as you mentioned is a generic name)
2) Marking every malware that uses `afraid.org` as a `sdbot` (Note: it doesn't too generic to me) doesn't make sense

stamparm:
As a compromise, if you filter out really those domains that have at least one occurrence of SDBot in VT, I'll merge it (that filtered version)

MikhailKasimov:
#933

MikhailKasimov:
This list was filtered three times to have explicit malicious trails and ```sdbot``` was the name, that was met _most of the times_ as **generic name** to all these tons of ```[a-z]*bot```-s: ```agobot```, ```gtbot```, ```blablabot```, ```lousy-shitty-crappybot``` and so on.

stamparm:
Done with 8c297ab

PR #950, opened 2019-03-16
Create zeroshell_scan.txt

MikhailKasimov:


stamparm:
Not worth a file on its own as scanners scan whatever its actual. Focusing on that front would just introduce too much junk

PR #953, opened 2019-03-16
Update ua.txt

MikhailKasimov:


stamparm:
`zmap` hasn't appeared in first reference (only fingerprint that could be based on anything)

PR #959, opened 2019-03-17
Update ua.txt

MikhailKasimov:


stamparm:
I've skipped for example `python` (default UA for python) because there are far too many appliance agents that use python for regular activities. I could merge this one, but I am afraid that there will be FP recognitions here (too)

MikhailKasimov:
Can you take some experiments on your side, plz? Or, if it is really risky to catch ton of FPs, no problem not to apply this trail.

stamparm:
I'll merge it and check the status in couple of days

MikhailKasimov:
OK, thnx!

PR #971, opened 2019-03-18
Update powershell_injector.txt

MikhailKasimov:


stamparm:
"Domain fronting" means that valid CDN domain (in this case Akamai's) is used to hide the real target (usually referenced through `Host`)

PR #982, opened 2019-03-18
Update generic.txt

MikhailKasimov:


stamparm:
```
# Reference: https://www.group-ib.com/media/js-sniffer/

gmo.li
```
-> `malicious_js.txt`

PR #1005, opened 2019-03-19
Create monerominer.txt

MikhailKasimov:
Moving to ```apt_unclassified.txt``` or renaming to ```apt_monerominer.txt``` is discussable.


stamparm:
I vote for `generic.txt` as A) it isn't APT related (I don't see an APT actor/story in the reference) and B) `monerominer` seems to be too generic to the other mining operations. It sounds like it contains all monero mining domains

MikhailKasimov:
OK, agreed.

MikhailKasimov:
Closed on #1013

PR #1006, opened 2019-03-19
Create filesharing_service.txt

MikhailKasimov:
Filesharing services can be considered 50% as clean and 50% as malicious. So, I propose as ```suspicious``` trail. Also current ```1fichier``` was mentioned as source of malware.


stamparm:
Creating this one will for sure go into wrong direction (filling of all file sharing service domains). For starter, I use file sharing services a lot for benign purposes (exchanging bulks of data)

PR #1061, opened 2019-03-21
Create apt_kabarcobra.txt

MikhailKasimov:


stamparm:
We already had a discussion related to the `apt_XXX`. `XXX` should be (APT) threat actor name, not operation. Also, trails are already inside `apt_sectora05.txt`

MikhailKasimov:
> Also, trails are already inside `apt_sectora05.txt`

Unexpected, but OK.

PR #1074, opened 2019-03-22
Update cobalt.txt

MikhailKasimov:


stamparm:
`cobalt.txt` is for Cobalt Gang. Cobalt Strike is not Cobalt Gang

MikhailKasimov:
> 
> 
> `cobalt.txt` is for Cobalt Gang. Cobalt Strike is not Cobalt Gang

I do not remember, which trail is responsible for Cobalt Strike... :( We have ```cobalt``` and ```apt_cobaltdickens```.

stamparm:
There is only a feed `cobaltstrike.py`. There are no static trails for this as `Cobalt Strike` is not a malware - it is an attacking framework.

As I can see some interesting domains, you could create `cobaltstrike.txt` for this case

MikhailKasimov:
> 
> 
> There is only a feed `cobaltstrike.py`. There are no static trails for this as `Cobalt Strike` is not a malware - it is an attacking framework.
> 
> As I can see some interesting domains, you could create `cobaltstrike.txt` for this case

Done in #1101

PR #1101, opened 2019-03-23
Create cobaltstrike.txt

MikhailKasimov:

PR #1121, opened 2019-03-25
Update cobalt.txt

MikhailKasimov:
Moving trails from #1101


stamparm:
cobalt != cobaltstrike

MikhailKasimov:
Ok, will also move it to ```generic.txt```

PR #1127, opened 2019-03-25
Create winrar_cve-2018-20250.txt

MikhailKasimov:
I propose explicit-named trail instead of ```generic.txt``` one, because it would be useful and interesting for forensics goals, if someone in your LAN tries to use such services of malware/exploit compilation/creation.

### DIFF
--- a/trails/static/malware/winrar_cve-2018-20250.txt
+++ b/trails/static/malware/winrar_cve-2018-20250.txt
@@ -1,0 +1,6 @@
+# Copyright (c) 2014-2019 Maltrail developers (https://github.com/stamparm/maltrail/)
+# See the file 'LICENSE' for copying permission
+
+# Reference: https://twitter.com/PhysicalDrive0/status/1109871662890127360
+
+lofi.stream


stamparm:
Doesn't deserve a trail file on its own. Also, nothing to do with `malware`. Really not sure where to put it

MikhailKasimov:
> 
> 
> Doesn't deserve a trail file on its own. Also, nothing to do with `malware`. Really not sure where to put it

Let us have it as ```suspicious``` with the same name (winrar_cve-2018-20250.txt). M?

stamparm:
There are gazillion sites with similar "malicious" purposes. Creating a separate CVE really doesn't make any sense. At most, you can put it inside the `malicious_php`.

One more thing. Chance that the local user will access such site are slim to none. Malicious user on his own network will do that and use it against the local user.

MikhailKasimov:
> 
> 
> There are gazillion sites with similar "malicious" purposes. Creating a separate CVE really doesn't make any sense. At most, you can put it inside the `malicious_php`.
> 
> One more thing. Chance that the local user will access such site are slim to none. Malicious user on his own network will do that and use it against the local user.

It would be intesting to check, if such user sits in my LAN and then to ask him a question, which would be started with words ```WTF are you doing...?```. Looking from that side only I would like to keep initial trail-name.

stamparm:
-> `suspicious/malicious_service.txt`

MikhailKasimov:
> 
> 
> -> `suspicious/malicious_service.txt`

OK. Closed via #1153

PR #1129, opened 2019-03-25
Create nokia_7plus.txt

MikhailKasimov:


stamparm:
Ton of sites do this. Chinese are only ones that don't hide that they do it. Not worth the trail file.

MikhailKasimov:
OK.

PR #1135, opened 2019-03-25
Update ipinfo.txt

MikhailKasimov:


stamparm:
`iplogger.org` is an IP tracker, not `ipinfo` service (where malware takes IP of caller)

PR #1184, opened 2019-03-27
Create apt_iec.txt

MikhailKasimov:
```iec``` is the generic neutral name -- personally, I do not want to move this to ```unclassified```.


stamparm:
-> `apt_unclassified.txt` (although, apt_unknown.txt was better - unclassified means that classification couldn't be done, not that the actors are unknown)

MikhailKasimov:
See #1187 .

P.S.
```although, apt_unknown.txt was better - unclassified means that classification couldn't be done, not that the actors are unknown```

As for me, it is synonimcal -> we can't give classification to APT, because actor in known. And ```unclassified``` sounds stronger, than helpless ```unknown```.  Deal of taste.

PR #1334, opened 2019-04-09
Create malicious_hta.txt

MikhailKasimov:
Similar logic to ```malicious_{js|php}.txt``` trails and as an additional tool for #1268


stamparm:
Redundant (as said in previous comment), but merging nevertheless

PR #1353, opened 2019-04-09
Create malicious_powershell.txt

MikhailKasimov:
As an additional tool for #1352 . Logic is entirely the same to #1334 .


stamparm:
p.s. I believe that this one redundant (because of general .ps1 coverage), but merged whatsover

PR #1365, opened 2019-04-12
Update malicious_service.txt

MikhailKasimov:
RU-lang site with info on how to hack IP-cameras.


stamparm:
Actually, providing (also) (sub)domain service

PR #1372, opened 2019-04-12
Create locked_ransomware.txt

MikhailKasimov:
Feel free to rename, if needed. I used the similarity to ```fbi_ransomware.txt``` naming scheme. ```locked``` here -- is suffix of encrypted files by this sample.


stamparm:
`fbi_ransomware` is just called by that name because the plain `fbi` would sound kind of WTF. All other ransomware is called like the rest of malware - by its name. In this case `locked` would do

PR #1399, opened 2019-04-14
Create atilla.txt

MikhailKasimov:


stamparm:
This domain doesn't look like C&C. Even in referenced sample contacted domain is different

MikhailKasimov:
I decided to put this as ```malware``` instead of ```malicious_service``` or ```malicious_forum```, because of promoting one only malware by ```atilla.vip```.

stamparm:
Frankly, I don't see a difference of this and `lofi.stream` inside `malicious_service`. Also, if visitor goes to `atilla.vip` that can't be marked the same as a regular malware contacting C&C

MikhailKasimov:
OK, will move this to ```_service.txt```

stamparm:
`961api.com` to me looks like a valid "licensing" platform, most probably used by "atilla". Not a C&C

MikhailKasimov:
OK, let's close this on this moment. Will create another instance, if something new is in public about this trojan.

PR #1417, opened 2019-04-15
Create naver.txt

MikhailKasimov:


MikhailKasimov:
It is phishing instead. Moved to https://github.com/stamparm/maltrail/pull/1420

PR #1446, opened 2019-04-17
Update apt_kimsuky.txt

MikhailKasimov:


stamparm:
p.s. so many ip trails. I'll really need to do some periodic trimming

PR #1461, opened 2019-04-17
Update malicious_js.txt

MikhailKasimov:


stamparm:
1) `malicious_js.txt` is for malicious JS (e.g. fooling with client UI or doing malicious redirects)
2) This list (you posted) is a list of web shells at compromised sites. Nothing to do with `malicious_js.txt` purpose. Also, those same "links" are in 99% dead/cleaned. There is really no reason to put the here. They have been used (long before made public) by attackers and that's it. Maltrail is not for storing such trails

MikhailKasimov:
OK

PR #1507, opened 2019-04-19
Update malicious_php.txt

MikhailKasimov:


stamparm:
Google says that there could be FPs

PR #1542, opened 2019-04-24
Create cobaltstrike.txt

MikhailKasimov:


stamparm:
Case specific. This would be the only (not good enough) trails to connect to cobalt strike.

MikhailKasimov:
@stamparm  Couldn't see other way how to detect this case. ```/TRAINING-BEACON/``` seems the constant snippet here. Just like ```/is-ready``` for ```houdini.txt```. If you have an idea how to detect this -- let's discuss.

PR #1554, opened 2019-04-24
Update elf_gafgyt.txt

MikhailKasimov:


stamparm:
I can't find those dirs/files in referenced link (nor pastebin). Also, even in case of finding them, I am not a fan of merging those two

PR #1590, opened 2019-04-26
Create cobaltstrike.txt

MikhailKasimov:
One more attempt to "make ```cobaltstrike.txt``` great again" (c). Also keep trying not to lose a hope on #1542 resurrection.


stamparm:
Please put that domain into `generic.txt`. Cobalt Strike is a junk software used by clicky "hackers" on their clicky "missions". I have already said that I won't make a trail file based on one random case, especially for this software that I have zero respect for

PR #1605, opened 2019-04-26
Update infinityrat.txt

MikhailKasimov:


MikhailKasimov:
```blackbook``` covers it.

PR #1642, opened 2019-04-28
Update ipinfo.txt

MikhailKasimov:
See details in #1641


stamparm:
Only for general purpose (not malware related) services. Also, as said, redundant

PR #1669, opened 2019-04-29
Update elf_generic.txt

MikhailKasimov:
IP is currently alive.

```clean``` is sh-downloader. Downloads Perl IRCBot.


stamparm:
These IPs are disposable and not worth of including into static trails. Already covering TONS of similar with `urlhaus` feed. Please take a look into https://urlhaus.abuse.ch/downloads/text/

MikhailKasimov:
OK, no problem.

PR #1702, opened 2019-04-30
Create CVE-2019-2725.txt

MikhailKasimov:


stamparm:
1) We discussed previously. Per-CVE trail files won't be merged. It will just create mess
2) There are other projects that handle CVE detection (e.g. Snort, Suricata, ET, etc.) on a lower level than Maltrail

MikhailKasimov:
@stamparm CVE here is just a trail-name. Let's have another name, if proposed paths are acceptable itselves, or let's move this to ```generic```. If proposed paths are not acceptable, ok, let's have this closed and let Snort, Suricata, ET handle this.

PR #1715, opened 2019-05-01
Update apt_chafer.txt

MikhailKasimov:
IPs are partially alive.


stamparm:
p.s. I'll soon start cleaning all IP trails older than (e.g.) month or two

PR #1716, opened 2019-05-01
Create rogue_rdp.txt

MikhailKasimov:
Also need an advice on https://twitter.com/malwaremustd1e/status/1075395164980400129


stamparm:
-> `generic.txt`. Not worth a separate trail file

stamparm:
p.s. as of advice, attacker trails are already being collected through feeds that are specialized in collecting and updating such lists

PR #1799, opened 2019-05-07
Update malicious_php.txt

MikhailKasimov:
Not sure what exect malware family does this trail belong to, Google says nothing wrong on having such trail as ```suspicious```. Potentially can be moved to ```malware``` trails.


stamparm:
Seems as a proper place

PR #1801, opened 2019-05-07
Update malicious_с2panel.txt

MikhailKasimov:


stamparm:
Shouldn't `/gate.php` be part of malware trails, not `malicious_c2panel`? As I recall, we introduced this file for malicious operators (not victims).

PR #1803, opened 2019-05-07
Update ua.txt

MikhailKasimov:


stamparm:
Not sure that default language (Go) UAs should be "suspicious". I am seeing those at regular basis from regular programs

MikhailKasimov:
OK

PR #1860, opened 2019-05-10
Update generic.txt

MikhailKasimov:


stamparm:
A) `silverterrier` stuff should go to `silverterrier.txt`
B) I have a feeling that there will be FPs in there as I can see that there are tons of domains. Even the [report](https://www.paloaltonetworks.com/apps/pan/public/downloadResource?pagePath=/content/pan/en_US/resources/whitepapers/unit42-silverterrier-rise-of-nigerian-business-email-compromise) says to expect FPs (e.g. `considerable number of actors run legitimate businesses and simply use malware to supplement their income` and `...our analysis also reveals that a much smaller subset (1.2%) of the domains were registered to support legitimate business functions...`)

MikhailKasimov:
So, you put in trash ``all my other work``, instead of just cutting off Palo Alto's trails only??

MikhailKasimov:
#1894

PR #1894, opened 2019-05-13
Update generic.txt

MikhailKasimov:
Fixing someone's ```...``` disrespective closing of entire trail instead of cutting off irrelevant records.

### DIFF
--- a/trails/static/malware/generic.txt
+++ b/trails/static/malware/generic.txt
@@ -3150,3 +3150,7 @@ ttreface.tk
 # Reference: https://twitter.com/malware_traffic/status/1128019457966735360
 
 dhlexpress.club
+
+# Reference: https://twitter.com/ActorExpose/status/1128018026673131521
+
+double-minded-elect.000webhostapp.com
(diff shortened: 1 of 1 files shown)


stamparm:
`disrespective closing of entire trail instead of cutting off irrelevant records` <- but there was more than couple of thousands of trails??? I thought that it was far more easier for you to do it

MikhailKasimov:
@stamparm Please, let's just cut off those snippets, that are irrelevant only. E.g: ```Mikhail, cut off the problematic records and then I'll merge other records you proposed for this trail```. -> this will save a lot of time for me not to re-parse and not to re-create records, which are good to be merged. Hence, such approach will save me a tone of nerves and time. Thanks!

stamparm:
If you go back in history for thousands of merge requests you'll see that I don't do "disrespective" rejects. In this case I thought that it is easier for you to recreate two separate merge requests after I close this one. Recreation of records would be easy peasy with [commit history](https://github.com/stamparm/maltrail/pull/1894/commits)

MikhailKasimov:
OK, sorry for being angry.

PR #1913, opened 2019-05-14
Update domain.txt

MikhailKasimov:
Almost all trails in #1907 use dynamic domains, but I'm not strongly sure, that ```5wya.com``` is dynamic one, because there's no much info about this domain. Due to graphs on VT, this domain leads to CN and HK only. And this domain is marked as one of actors-participants of attack in ```ptsecurity``` survey. So, let him live in ```domain.txt``` trail.

UPD: Or even to place all these proposed domains to ```bad_history.txt``` trail...

### DIFF
--- a/trails/static/suspicious/domain.txt
+++ b/trails/static/suspicious/domain.txt
@@ -4959,6 +4959,6 @@ zik.dj
 
 iok.la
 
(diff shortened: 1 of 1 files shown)


stamparm:
p.s. in case of "dynamic domain" there should be lots of subdomains in https://www.virustotal.com/#/domain/5wya.com

PR #1948, opened 2019-05-17
Create dirtjumper.txt

MikhailKasimov:
@stamparm Need also an advice how to proceed UA list of this malware: https://pastebin.com/4BD1PAcw


stamparm:
Majority of these UA are junk OR simulating some other (ro)bot. I wouldn't spend too much time there.

MikhailKasimov:
OK, won't be added.

PR #1952, opened 2019-05-18
Add profeehost.com to free_web_hosting.txt

youdly:


youdly:
There are many duplicate domains, should it be deduplicated?
```sh
$ cat trails/static/suspicious/free_web_hosting.txt | sort | uniq -c | sort -rn | sed 's/^ *//g' | head -10
4 net46.net
3 greatnow.com
3 freewebsitehosting.com
3 freesite.org
3 exactpages.com
3 envy.nu
3 dreamstation.com
3 digitalzones.com
3 builtfree.org
```

stamparm:
@youdly no need as Maltrail does the de-duplication automatically

PR #2017, opened 2019-05-24
Update crypto_mining.txt

MikhailKasimov:


stamparm:
Exchanges, not mining pools

PR #2137, opened 2019-06-01
Create mylobot.txt

MikhailKasimov:


stamparm:
Using DGA for static trails doesn't make much sense

MikhailKasimov:
Better, than nothing...

PR #2139, opened 2019-06-01
Create crylocker.txt

MikhailKasimov:


MikhailKasimov:
@stamparm Some doubts on merging?

stamparm:
@MikhailKasimov yup. Not fan of merging thousands of IP:PORTs, while as far as I could understand, attacker is just hiding behind one of those

MikhailKasimov:
@stamparm 

OK, my point of view:

- As I can see, this list is related to sample. Sample is old, but there's no garantee of its resurrection under undetectable packer somewhere and this list potentially can be actual. It is a little bit paranoic, but I prefer to give user detection instead of missing.

- On moment of this sample was active, MT didn't have functionality of IP:PORT (this was implemented just on Sep 1 2018: https://github.com/stamparm/maltrail/commit/a00d600ea20a4bced54d84f12034972e01d5678b). So, it couldn't be added on its being active. Detection of this can be considered as a bug-fix.

- Port (4444) is specific (similar to merged https://github.com/stamparm/maltrail/pull/2150/commits/f4cb4416df131511a6a06cc16a9f0c1c89d12faf for port 4443) and I do not expect FPs on here.

Hope, this looks good.

stamparm:
Its not FP that worries me. It is the inclusion of 76KB (almost 100KB) of IP:PORT lines because of (potentially) one IP:PORT where attacker is hiding. I mean, only one line inside is potentially valuable, while the rest is pure noise (according to the report you referenced).

In plain speak, cost-benefit is pretty low from my point of view.

MikhailKasimov:
OK, can we have something like ```0.0.0.0:4444 crylocker (heuristic)``` for such cases?

Generally, I was thinking on having some port-list like https://www.garykessler.net/library/bad_ports.html (for TCP ports >=1024 only). E.g. for ```njrat```: https://www.speedguide.net/port.php?port=1177

Applicable?

stamparm:
I have a strange feeling that something could go wrong when blacklisting ports. I could do it, but that should be used ONLY when >99% sure that malware is behind it. Putting all sorts of junk (e.g. for prehistoric malware) here would do more harm than benefit

MikhailKasimov:
OK, while thinking on this fucntionality, I can cut problematic snippet off.

MikhailKasimov:
@stamparm IPs list is (temporary?) dropped. Merging?

MikhailKasimov:
@stamparm ?

PR #2184, opened 2019-06-04
Update domain.txt

MikhailKasimov:


stamparm:
If domains don't have any other subdomains than in malware itself, then it should be inside a malware trail, not suspicious `domain`

MikhailKasimov:
OK, will move them all to ```generic.txt```

MikhailKasimov:
Closed via https://github.com/stamparm/maltrail/pull/2205

PR #2197, opened 2019-06-04
Update powershell_injector.txt

MikhailKasimov:


stamparm:
Invalid address (private LAN range)

PR #2216, opened 2019-06-06
Create petya.txt

MikhailKasimov:
Very strange how this could be missed... o_O


stamparm:
Basically, those onion domains are official C&Cs. In regular situations, those would be contacted through Tor, thus invisible to Maltrail. That's the reason for skipping

PR #2232, opened 2019-06-07
Create android_coinminer.txt

MikhailKasimov:


stamparm:
-> `android_generic` as this malware does not do coin mining (as I could see from references)

MikhailKasimov:
Resounded in 2239

PR #2238, opened 2019-06-08
Update crypto_mining.txt

MikhailKasimov:


stamparm:
I've stated numerous times before that paying is not the same as mining

PR #2256, opened 2019-06-10
Create ghost419.txt

MikhailKasimov:
```Ghost419 is based on Gold Dragon and Brave Prince implants and contains shared elements and code, especially for system reconnaissance functions.```

### DIFF
--- a/trails/static/malware/ghost419.txt
+++ b/trails/static/malware/ghost419.txt
@@ -1,0 +1,17 @@
+# Copyright (c) 2014-2019 Maltrail developers (https://github.com/stamparm/maltrail/)
+# See the file 'LICENSE' for copying permission
+
+# Aliases: brave prince, gold dragon, ghost419
+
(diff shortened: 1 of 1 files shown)


stamparm:
-> `apt_unclassified.txt`. AFAIK, those are tool/malware names used by unknown actors

MikhailKasimov:
Mmmm... I'm standing for some duplication: 1) we have malware-named trail + 2) apt, where this malware took is/was a participant. So, ```ghost419.txt``` + ```apt_unclassified.txt```. M?

stamparm:
Golden rule should be: if malware is used ONLY in case of APT attacks by particular group, then preferred way should be to put it in `apt_XYZ` trail. In all other cases there could be two trail files.

Just ask a question to yourself. If `foobar.com` domain is being found inside the DNS traffic, should user be warned of APT group of malware name. I believe that APT group should have a precedence as it seems more important in matter of things

MikhailKasimov:
OK, one minute...

MikhailKasimov:
Resounded in https://github.com/stamparm/maltrail/pull/2265

PR #2286, opened 2019-06-12
Create elf_hacked_exim.txt

MikhailKasimov:

PR #2298, opened 2019-06-13
Update elf_generic.txt

MikhailKasimov:
Working on ```elf_hacked_exim.txt``` trail (#2286 ), have found bash-loader ```173.212.214.137/b``` with addresses, which can be marked as ```elf_generic.txt```.


stamparm:
These will disappear in a week. Putting all those IPs from (e.g.) Mirai download places would result with a total mess

MikhailKasimov:
I really want to have some different way to detect this, but there is no better way to do it, because files to download have ```/b```, ```/s```, ```/se``` names and there's no domain names. But somehow this should be detected and current IP is not in use for something normal (site, mail, etc). Sorry.

PR #2356, opened 2019-06-18
Update agenttesla.txt

MikhailKasimov:


stamparm:
Doesn't look like C&C. Most probably sent from that same address. Also, name looks like a generic machine name for a shared hosting provider (`kvchosting.com`)

stamparm:
p.s. there is no such DNS request inside https://app.any.run/tasks/dbbdbf98-4153-482e-bf06-2e291f54626d/

PR #2415, opened 2019-06-21
Create loudminer.txt

MikhailKasimov:


MikhailKasimov:
Mining hosts are in #2416

PR #2469, opened 2019-06-26
Update ua.txt

MikhailKasimov:
Can't see this UA in trail's history and declining this as FP generator. Please, re-check.


stamparm:
https://github.com/kennethreitz/grequests  # regular library

MikhailKasimov:
Ah, ok! Thanks!

PR #2478, opened 2019-06-27
Update agenttesla.txt

MikhailKasimov:


stamparm:
Doesn't have anything to do with AgentTesla. Plain SMTP server

PR #2508, opened 2019-06-28
Update luminositylinkrat.txt

MikhailKasimov:


stamparm:
Ok about domains, but IPs are from 2015-2016. Please only domains

MikhailKasimov:
IPs have thier specific ports. In case of 80, 443 -- OK, but when specific ports are met, standing to save the detection.

stamparm:
3 years old IOCs that are not domain names are unusable at least (I don't want to use other wording)

MikhailKasimov:
@stamparm https://github.com/stamparm/maltrail/pull/2508/commits/533090edd4be8e426a927da42bbed4d5e76764f2

PR #2520, opened 2019-06-29
Update malicious_js.txt

MikhailKasimov:


stamparm:
This one doesn't look convincing. Actors used randomly picked names for JS, where some of those names look like they could provoke FP detections

MikhailKasimov:
50/50. This trail was proposed for additional ensurance. No problem not to use it.

PR #2563, opened 2019-07-01
Delete dprk.txt

MikhailKasimov:
That was test trail.


MikhailKasimov:
Closed. Needed as an example.

PR #2655, opened 2019-07-06
Update generic.txt

MikhailKasimov:
Moving to ```elf_generic.txt``` : #2654


stamparm:
Please, I can't merge this kind of stuff. This means that every DGA-alike malware could potentially introduce vast amount of domains. I'll do some support for regexes, but this kind of "quick" way-around doesn't make any sense. Memory and (disk) space consumption would just explode

MikhailKasimov:
OK, I'll remove ```helplog*```.

```I'll do some support for regexes``` <-- this will really appreciated. Thanks!

MikhailKasimov:
@stamparm Done with https://github.com/stamparm/maltrail/pull/2655/commits/281a4d7609c4ef9a165de5acb94e9cb92b8f32a4

MikhailKasimov:
@stamparm BTW, ```some support for regexes``` will also really be apprecicated for ```android_roamingmantis``` trails (aa-.AA.TLD, jppost-AA.TLD) + some Japanese spam, which use currently the same to ```android_roamingmantis``` naming scheme for their domains.

PR #2683, opened 2019-07-08
Update trickbot.txt

MikhailKasimov:


stamparm:
Too many FPs here. Majority of these look like compromised

MikhailKasimov:
Oops, sorry. Then closing.

PR #2713, opened 2019-07-09
Update generic.txt

MikhailKasimov:


stamparm:
This is a good example of what we spoke earlier:

For example, this one:
```
anothertembr.cf
anothertembr.ga
anothertembr.gq
anothertembr.ml
```

should be more preferred than this one:
`\banothertembr\.(cf|ga|gq|ml)`

as table lookups are least expensive. I'll just then put a kind of best practice rule: use regular expressions in case of >= 10 domains sets. For everything else, use raw entries

MikhailKasimov:
OK, rule is accepted.

PR #2752, opened 2019-07-11
Update mass_scanner.txt

MikhailKasimov:


stamparm:
Looks like a temporary IP (VPS) used for mass scans. Please use this one for "established" mass scanners (as are other listed here)

PR #2753, opened 2019-07-11
Update magentocore.txt

MikhailKasimov:


stamparm:
Majority are compromised sites (while you shortened IOCs to domain names)

PR #2763, opened 2019-07-11
Update generic.txt

MikhailKasimov:

PR #2766, opened 2019-07-11
Update android_roamingmantis.txt

MikhailKasimov:
To avoid potential detection in the middle of some strings/domains/etc.


stamparm:
https://github.com/stamparm/maltrail/pull/2763#issuecomment-510829734

MikhailKasimov:
@stamparm Fixed

stamparm:
@MikhailKasimov p.s. just be aware that `^` will skip cases like `www.<malicious_domain`. So, use it wisely

MikhailKasimov:
> 
> 
> @MikhailKasimov p.s. just be aware that `^` will skip cases like `www.<malicious_domain`. So, use it wisely

Yeah, I already thought on it when adding today's regex for phishing. On ```roamingmantis``` there are no domains with ```www.```.

PR #2793, opened 2019-07-13
Update gandcrab.txt

MikhailKasimov:


stamparm:
Majority are regular domains. Also, on that same reference page: `Domain Names contacted by malware. Does not indicate maliciousness`

PR #2834, opened 2019-07-15
Create ad_network.txt

MikhailKasimov:
Reason of creating this trail: ```RouterCSRF campaigns are often distributed via malvertising campaigns through popcash[.]net, dolohen[.]com, rotumal[.]com, bodelen[.]com and other ad rotators, and appear in waves.```


stamparm:
This basically means that there will be LOTS of "suspicious" DNS requests also for a regular ad traffic

MikhailKasimov:
Even on mentioned in article?

stamparm:
There are tens of similar cases where malware has been distributed via ad networks. Even via Google, Yahoo and MSN. This is really not a good idea, no matter what the "article" says

MikhailKasimov:
OK...

PR #2836, opened 2019-07-15
Create socks5_proxy.txt

MikhailKasimov:


stamparm:
This could really go out of control. Copy-pasting SOCKS proxies here IMO will be a disaster. SOCKS proxies come and go. Malware can use any SOCKS proxy it wants

MikhailKasimov:
OK, closing.

Issue #2140, opened 2019-06-01
[Phishing.Database] Feeds for consideration


MikhailKasimov:
Hello!

Subj: https://github.com/mitchellkrogza/Phishing.Database/tree/master/phishing-domains/output/domains

MikhailKasimov:
ping -c1 @stamparm

MikhailKasimov:
All modern browsers gonna save the world: https://github.com/stamparm/maltrail/pull/3223

PR #2853, opened 2019-07-16
Update phishing.txt

MikhailKasimov:


stamparm:
Rationale behind listing all domains for "185.28.193.195"?

MikhailKasimov:
Alternatives?

stamparm:
I am only wondering why have you picked that same IP (185.28.193.195)?

MikhailKasimov:
Initially I checked links from https://twitter.com/521_maika/status/1077019453957668866 and then re-checked domains on VT, after then I checked the IP and then... picked up all other domains. NS-servers went to ```rogue_dns.txt``` (https://github.com/stamparm/maltrail/pull/2869/commits/5996a0996a8751305bad07be97eeb4b67ee88d8c)

stamparm:
As far as I remember, `rogue_dns` was created for "DNS hijacking" cases. Now you are telling me that also "phishing" related name servers are going there too?

stamparm:
Anyway, I would leave `rogue_dns` for DNS hijacking cases, as originally intended, while putting those "problematic" name servers into `phishing`.

MikhailKasimov:
50/50: if someone will change (hijack) DNS to ```phishing``` DNS, what is the difference? Phishing DNS is also an rogue. I can't see big difference here...

stamparm:
I can see this going wrongly in so many places. I've warned multiple times in previous occasions to not focus too much on "phishing". Browsers nowadays have a lightning-fast propagation of such domains. Also, Maltrail uses `openphish` feed which should be more than enough.

Marking tens of thousands of domains as "phishing" and DNS servers of their registrants as "rogue" won't give any added value to this project AS that fight is being fought at different fronts (e.g. browsers as mentioned).

Malware domains have a value. Malware is reminant in local networks and generates (e.g. DNS) trails for detection.

Phishing domains have a short (1 day) usability window - while browsers are doing a really good (ACTIVE vs Maltrail's PASSIVE) job in preventing access to those kind of domains

stamparm:
p.s. https://github.com/stamparm/maltrail/commit/034891dbe43f3b7142eb46e1008b56006f7b6f46

rogue

MikhailKasimov:
> I can see this going wrongly in so many places. I've warned multiple times in previous occasions to not focus too much on "phishing". Browsers nowadays have a lightning-fast propagation of such domains. 

Not all the time browsers highlights phishing, unfortunately (latency). That is the reason why I want to decrease this latency: if MT is installed on some gate, then to have the detection without concentrating on endpoint browsers.

>Also, Maltrail uses `openphish` feed which should be more than enough.

And #2140 , which is currently wiwth no reply.
> 
> Marking tens of thousands of domains as "phishing" and DNS servers of their registrants as "rogue" won't give any added value to this project AS that fight is being fought at different fronts (e.g. browsers as mentioned).
> 
> Malware domains have a value. Malware is reminant in local networks and generates (e.g. DNS) trails for detection.
> 
> Phishing domains have a short (1 day) usability window - while browsers are doing a really good (ACTIVE vs Maltrail's PASSIVE) job in preventing access to those kind of domains

Generally -- no problem at all to move DNS-trails to phishing. Yes, initially ```rogue_dns``` was designed to have detection for cases of hijacked routers, but I try to look wider and cover all cases of suspicious/malicious/phishing DNS-es regardless on are they a part of some real attack or potential attack. Again, no problem at all to bring phishing DNS-es from ```rogue_dns``` to ```phishing```. If yes, just put a phrase like ```yes, please, move them to phishing.```.

PR #3001, opened 2019-07-25
Add CNCERT/CC sinkhole IP

wesinator:


stamparm:
@wesinator looks like sinkhole, though, I can't find any info for it to be related to CNCERT. Can you just please pinpoint how did you connect that IP with CNCERT?

wesinator:
> @wesinator looks like sinkhole, though, I can't find any info for it to be related to CNCERT. Can you just please pinpoint how did you connect that IP with CNCERT?

http://183.236.2.18:8080 :)

image

stamparm:
Fair enough :)

PR #3027, opened 2019-07-27
Update ua.txt

MikhailKasimov:


stamparm:
I'll merge this, but I expect FPs here

MikhailKasimov:
Hope, they won't appear. It even will be interesting what exactly should trigger as FP.

PR #3129, opened 2019-08-04
Update domain.txt

MikhailKasimov:
From https://github.com/stamparm/maltrail/pull/3096/commits/1c09a821212cd5d9fa53b40fa301e11c29c9d363


stamparm:
Those are only DNS registrars. Client contacting them doesn't prove anything else than that the original domain is registered there (while domain can be anything). For example, `dnspod` is known to hold lots of scum domains, but GoDaddy is even worse.

PR #3163, opened 2019-08-06
Update malicious_js.txt

MikhailKasimov:


stamparm:
Same named sorting (non-malicious) JS

PR #3173, opened 2019-08-07
Update ua.txt

MikhailKasimov:


stamparm:
I have a feeling that this could cause FPs (regular Windows UA)

PR #3183, opened 2019-08-08
Update phishing.txt

MikhailKasimov:


stamparm:
Marking whole `namecheaphosting` (`68.65.122.53`) as phishing is wrong

MikhailKasimov:
> 
> 
> Marking whole `namecheaphosting` (`68.65.122.53`) as phishing is wrong

Fixed.

stamparm:
I can see that you've started to force usage of regular expressions, especially in phishing cases. For example, can you please tell me valid examples (known phishing cases) for:

```
[a-z0-9-_.]{0,}amazonlimited[a-z0-9-_.]{0,}
[a-z0-9-_.]{0,}secur[a-z0-9-_.]{0,}\.(tk|gq|ga|ml|cf|icu|top|xyz)
ebaytay[a-z0-9-_.]{0,}\.(tk|gq|ga|ml|cf|icu|top|xyz)
```

stamparm:
p.s. I've already said for couple of times. Please use regular expressions only IF absolutely required. Too many regular expressions will slow down the whole processing

MikhailKasimov:
> 
> 
> I can see that you've started to force usage of regular expressions, especially in phishing cases. For example, can you please tell me valid examples (known phishing cases) for:
> 
> ```
> [a-z0-9-_.]{0,}amazonlimited[a-z0-9-_.]{0,}
> [a-z0-9-_.]{0,}secur[a-z0-9-_.]{0,}\.(tk|gq|ga|ml|cf|icu|top|xyz)
> ebaytay[a-z0-9-_.]{0,}\.(tk|gq|ga|ml|cf|icu|top|xyz)
> ```

These all from https://twitter.com/PhishFindR/status/1159815531148525569 nad are based on

```
amazonlimited.myvnc.com
ca-secure-09384.icu
ebaytay.tk
```

domains. This is an attempt to ```have extrapolation`` to detect something like ``` payment-secure-paypal.xyz```, ```secure1.verification.cf``` and so on, which are NEVER be legit by default. I never believe, that ```amazonlimited``` can be a sign of legit ```amazon.com```. And I never believe, that something, which is related to ```ebay```, can be in ```(tk|gq|ga|ml|cf|icu|top|xyz)``` domains.

Hence, again, this is an attempt to ```have extrapolation`` to detect ANY such phish, even if it is not appeared today.

> p.s. I've already said for couple of times. Please use regular expressions only IF absolutely required. >Too many regular expressions will slow down the whole processing

Attentive eye can notice, that I use regexes for phish domains, which can be relative to ```apple```, ```ebay```, ```amazon````, etc key-words. Or, when there are no variations to be legit (e.g. ```softbank.info```, ```wsoftbank.info```, ```ebay-some-mess.tk``` etc). In other cases I put trails AS IS.

stamparm:
1) we've already had multiple discussions related to phishing. Those are of least concern as all modern browsers have more than instant blacklist-based anti-phishing URL checks
2) please don't do extrapolations because in general case all domains (malware and non-malware) could be extrapolated. I've asked you multiple times to restrain usage of regular expressions as I don't want those to become a performance bottleneck. 

I just checked. Currently there are 324 regular expressions already included. Majority are `...apple...` something. Because of this I need to temporary restrain inclusion of any new regular expressions - as this can all go in a very bad direction

MikhailKasimov:
And all non-regex phish domains are also closed. Great.

PR #3215, opened 2019-08-10
Update whitelist.txt

MikhailKasimov:
To decrease potential FPs, when detecting phish, which targets these resources. Feel free to update this list.


stamparm:
Whitelist doesn't support regexes

stamparm:
There is no point in using regular expressions in the first place here. Those are simple domains. Also, please use only the main domain as all those subdomains can already be covered with main domains

MikhailKasimov:
OK

PR #3237, opened 2019-08-13
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
Please put it to `domain.txt`. I can see that those two providers mark it as "dynamic", but I believe that they marked it (most probably) because of its (potential) fast-flux capabilities.

All those currently point to 4 dummy IPs. Dynamic DNS providers are used in a totally different manner

MikhailKasimov:
OK. Closed via https://github.com/stamparm/maltrail/pull/3241

PR #3275, opened 2019-08-16
Update bogon_ranges.txt

MikhailKasimov:


stamparm:
https://en.wikipedia.org/wiki/Bogon_filtering

stamparm:
Ranges `185.169.230.0/24` and `185.169.231.0/24` are used for malicious purposes, but they are not "bogon" (as the original idea was for `bogon_ranges.txt`)

MikhailKasimov:
Hence, ```bad_packets``` info wrong is?

stamparm:
Nope. They are using "bogon" to generally define IP ranges which are "bad" (propagated at BGP level). `bogon` in Maltrail is used to define ranges which should not be used in the first place (private and reserved), where their appearance usually means something "suspicious" is going on.

`bad_packets` can mark something as `bogon`, while those same ranges can be JUST used for a temporary traffic (BGP) takeover - in month or two those same ranges will be used for (e.g.) VPS/VPN services

MikhailKasimov:
> bad_packets can mark something as bogon, while those same ranges can be JUST used for a temporary traffic (BGP) takeover - in month or two those same ranges will be used for (e.g.) VPS/VPN services

OK, in this case, we can bravely close this PR, I suppose, because this ```bad_packets``` info is dated with 2018. Right?

stamparm:
Yup

PR #3283, opened 2019-08-17
Update ipinfo.txt

MikhailKasimov:


stamparm:
Original purpose of `ipinfo.txt` is to have **specialized** ipinfo services (not some random pages with `<?php $_SERVER["REMOTE_ADDR"]; ?>` used in isolated cases)

MikhailKasimov:
OK, closed.

PR #3290, opened 2019-08-18
Create elf_backconnect.txt

MikhailKasimov:


stamparm:
Same thing as for shell_backconnect. This doesn't reserve a category on its own. Tons of malware are "backconnect", while these are really a crap

MikhailKasimov:
Closed via #3297

PR #3324, opened 2019-08-20
Create ruby_backdoor.txt

MikhailKasimov:


stamparm:
Just one thing. I would not like to merge this to later go into the discussion into why we should have generalization for this SPECIFIC case (for any other backdoor written in XYZ language)

MikhailKasimov:
Reason of current name: being inspired of ```elf_hacked_mint```. Faceless ```generic``` looks (personally) good, but doesn't show the nature of this malware. Also we have ```python_xwo.txt```. So, (again -- personally) I can see no something bad to have specific names for such cases.

stamparm:
I am clear of naming. I just urge to not come with a generalization (as in recent cases of reverse-shell malware) and put all ruby "backdoors" into here

MikhailKasimov:
Ah, OK.

PR #3334, opened 2019-08-21
Update rogue_dns.txt

MikhailKasimov:


stamparm:
Twitter reference says `?`. Removed IP is Amazon AWS located, hence, I would not like to risk FPs

PR #3408, opened 2019-08-27
Update dynamic_domain.txt

MikhailKasimov:
Weird, but due to https://www.virustotal.com/gui/domain/dyndnss.net/relations --  this doesn't seem to be a dyndns...


stamparm:
`Weird, but due to https://www.virustotal.com/gui/domain/dyndnss.net/relations -- this doesn't seem to be a dyndns...` <- looks like those are being redirected to the same IP where warned to be not updated periodically

PR #3557, opened 2019-09-05
Update generic.txt

MikhailKasimov:
See also https://pastebin.com/k3N36wMm


stamparm:
p.s. those `conn:` look "usable"

PR #3733, opened 2019-09-17
Update konni.txt

MikhailKasimov:


stamparm:
It is already there (lines 24 and 31)

PR #3743, opened 2019-09-17
Update emotet.txt

MikhailKasimov:


stamparm:
I believe that we've had related discussion previously. Please don't bother with temporary IP:PORT C&C addresses for Emotet. They come and go

MikhailKasimov:
I undertand that, but we also should somehow detect Emotet...

PR #3780, opened 2019-09-19
Update rogue_dns.txt

MikhailKasimov:
```37.59.52.229``` resolved on ```.xyz``` domains + ```podniks.com``` is marked as ```known infection source```. So, I think, we can have detection for ```37.59.52.229:53``` as ```rogue_dns```.


stamparm:
Doesn't look like rogue DNS. Rogue DNS should do "rogue" redirections for known domains/hostnames (e.g. www.google.com). In this case, this more looks like an authoritative server for `podniks.com` domain

MikhailKasimov:
OK

PR #3909, opened 2019-09-28
Update settings.py

MikhailKasimov:
```netsh``` + ```rm -(r|R)f``` commands to detect.


stamparm:
As said in previous related comment, please don't over-engineer this one as it is already sufficient for vast majority of cases.

By this logic we could introduce all linux and windows commands, while it doesn't make sense in the end to do it

MikhailKasimov:
OK

PR #3921, opened 2019-09-28
Create python_pyleet.txt

MikhailKasimov:


stamparm:
Changed name to not introduce one more subcategory (where it won't grow in size as others)

MikhailKasimov:
```python_xwo.txt```

PR #4016, opened 2019-10-04
Update malicious_js.txt

MikhailKasimov:


stamparm:
Those domains look like valid

MikhailKasimov:
> Those domains look like valid

Fixed in https://github.com/stamparm/maltrail/pull/4016/commits/04c08f35917f416a70cd9490edfcc84d3bb45db2

PR #4051, opened 2019-10-05
Create hacked_ashleymadison.txt

MikhailKasimov:


stamparm:
MX servers used for sending scam/spam emails. Long forgotten. There should be no malware or similar contacting those from local network nowadays

MikhailKasimov:
OK

PR #4054, opened 2019-10-05
Create ek_angler.txt

MikhailKasimov:


stamparm:
There is already `angler.txt`, which should be renamed to `ek_angler.txt` (to be inline with other `ek_XXX` trails)

MikhailKasimov:
Superseded by https://github.com/stamparm/maltrail/pull/4074

PR #4064, opened 2019-10-05
Create xtbl.txt

MikhailKasimov:


stamparm:
Please check domains coming from that same "advanced research" source. Lots of FPs

PR #4120, opened 2019-10-07
Update agenttesla.txt

MikhailKasimov:


stamparm:
Valid domain registered in 2014 (AgentTesla tends to use compromised servers/accounts)

PR #4144, opened 2019-10-08
Create cookiemonster.txt

MikhailKasimov:


stamparm:
-> `malicious_js`. As discussed previously, main problem here is the malicious JS (which can work OR not on the compromised sites). Hence, not a classical malware case

MikhailKasimov:
OK

MikhailKasimov:
Closed via https://github.com/stamparm/maltrail/pull/4157

PR #4262, opened 2019-10-14
Create apt_ta407.txt

MikhailKasimov:


stamparm:
`also known as Silent Librarian, Cobalt Dickens, and Mabna Institute`, while we already have apt_cobaltdickens`

MikhailKasimov:
OK, will move.

MikhailKasimov:
Resounded in #4268

PR #4347, opened 2019-10-17
Update magentocore.txt

MikhailKasimov:


stamparm:
Compromised site (registered in 2013)

PR #4396, opened 2019-10-19
Update lokibot.txt

MikhailKasimov:


stamparm:
`fueda.info` is already at https://github.com/stamparm/maltrail/blob/master/trails/static/malware/lokibot.txt#L1030

PR #4510, opened 2019-10-24
Update elf_mirai.txt

MikhailKasimov:


MikhailKasimov:
It is already present. Sorry for making noise.

PR #4546, opened 2019-10-27
Update dynamic_domain.txt

MikhailKasimov:
Belongs to ```afraid.org```.


MikhailKasimov:
Closed via #4547, which already contains this trail.

PR #4626, opened 2019-10-31
Update emotet.txt

MikhailKasimov:
Let it live here for a while. Will move to other, if something changes.


MikhailKasimov:
Closed via #4627

PR #4754, opened 2019-11-05
Update aspire.txt

MikhailKasimov:
New trail + some optimization.


stamparm:
p.s. with IPs it is better to leave those paths (to reduce FPs in future)

PR #4815, opened 2019-11-07
Update domain.txt

MikhailKasimov:
Many subdomain's names are rather weird to be legit.


stamparm:
Looks like regular hosting (https://hosting.miarroba.com/indice.php)

PR #4871, opened 2019-11-09
Create wp-vcd.txt

MikhailKasimov:


stamparm:
p.s. not classical malware, per-se. Hence, removed those "generic trails" as I expect those to be triggered more by scanners than by regular malware calls. Also, been in doubt whether to reject or merge this in the first place, but as those domains should be triggered only in case that web server is "infected", decided to merge those

PR #5002, opened 2019-11-17
Update ek_rig.txt

MikhailKasimov:


stamparm:
p.s. please regexes only if absolutely needed

PR #5086, opened 2019-11-21
Update apt_redoctober.txt

MikhailKasimov:


stamparm:
Something is funky behind this whole attribution:

https://www.bankinfosecurity.com/red-october-apt-gang-resurfaces-a-7667

f

p.s. I have an urge to split red october and cloud atlas as it does not make any sense that red october started targeting russian entities

stamparm:
p.p.s. split with the latest commit. Red october had totally different targets. This attribution/connection stinks as much as "Lazarus"/North Korea from start

PR #5113, opened 2019-11-22
Create apt_sandworm.txt

MikhailKasimov:


stamparm:
actually same team as `apt_blackenergy`. will merge and "patch"

PR #5229, opened 2019-11-28
Update konni.txt

MikhailKasimov:


stamparm:
FP (tweet also says "joke")

PR #5322, opened 2019-12-02
Update vidar.txt

MikhailKasimov:


stamparm:
` bitbucket.org`??

MikhailKasimov:
As I can see -- yes. But it is compromised legal site.

stamparm:
`But it is compromised legal site` <- I doubt. Most probably left something maliciously inside a personal repository

PR #5521, opened 2019-12-11
Create boteye.txt

MikhailKasimov:


stamparm:
1) not malware. API call from a bot's PHP
2) could be put inside the "malicious service", but AFAIK that IP is non-working

PR #5616, opened 2019-12-15
Add Sedo parking IP 91.195.240.94

wesinator:
is the whole 91.195.240.0/24 sedo ?


stamparm:
Well, yes. I've just put IPs as I've found domains pointing to them

PR #5669, opened 2019-12-18
Create mailerbot.txt

MikhailKasimov:


stamparm:
-> `generic.txt` (he used too generic name)

PR #5892, opened 2020-01-01
Update limerat.txt

MikhailKasimov:


MikhailKasimov:
Moving to ```quasarrat.txt```

PR #6061, opened 2020-01-09
Update lokibot.txt

MikhailKasimov:
Removing dup (see line 1535).


stamparm:
I know/knew that there is a duplicate. Just wanted to enlist that it appeared at that source too. The sensor itself deduplicates this kind of stuff, so there is no need to be strict here

PR #6104, opened 2020-01-12
Update magentocore.txt

MikhailKasimov:
Missed trail to detect. Other stuff is already present.


MikhailKasimov:
Closed as being a compromised site.

PR #6133, opened 2020-01-14
Update apt_unclassified.txt

MikhailKasimov:
This looks as TA505-based domain names structure, but I'm not really sure it is TA505 in fact. Or FacyBear, or anything else. While being with no explicit classification, I'm gonna put such cases to ```apt_unclassified.txt```. Also moving here the similar domain from ```domain.txt``` trail.


MikhailKasimov:
Resounded in #6138

PR #6282, opened 2020-01-22
Update emotet.txt

MikhailKasimov:
Tails can be various.


stamparm:
As said months ago, if trail can be more specific than the IP itself, it has a greater value because in future those same IPs will be used by valid web services by some other entity. Also, emotet gang doesn't use those same IPs for long, thus, there are not many paths per IP to look for

MikhailKasimov:
Reverting all back in https://github.com/stamparm/maltrail/pull/6282/commits/788e690f28d3af436861a9d66b398e3dc685629f

Issue #6297, opened 2020-01-23
PyPI


jayvdb:
It would be very helpful to be able to use maltrail as a library as part of other apps.

The name is currently available on PyPI https://pypi.org/project/maltrail/ , and even if there is no interest in regularly publishing on PyPI it would be sensible to reserve that name on PyPI to prevent someone else taking it.

This could be done without much restructuring I believe, with `trails` included as-is, and with `core` mapped to a library package called `maltrail`.  That would be more than sufficient to make the package usable as part of other projects.

The main change I see being needed is use of `thirdparty` should be changed so that `thirdparty` is inserted into `sys.path` in the scripts, and then `odict` and `six` loaded as normal packages.

jayvdb:
There is at least one other app which is using maltrail as a component https://github.com/GlacierSheep/DomainBlockList/tree/master/craw/modules/trail

jayvdb:
Some more copy&pastes https://github.com/search?utf8=%E2%9C%93&q=suspendedpage.cgi+language%3APython&type=Code&ref=advsearch&l=Python&l=

PR #6298, opened 2020-01-23
Make core and trails installable

jayvdb:
Implements https://github.com/stamparm/maltrail/issues/6297

### DIFF
--- a/core/common.py
+++ b/core/common.py
@@ -32,8 +32,8 @@ from core.settings import WHITELIST
 from core.settings import WHITELIST_RANGES
 from core.settings import WORST_ASNS
 from core.trailsdict import TrailsDict
-from thirdparty import six
-from thirdparty.six.moves import urllib as _urllib
+import six
+from six.moves import urllib as _urllib
 
 _ipcat_cache = {}
 
--- a/core/datatype.py
+++ b/core/datatype.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from thirdparty.odict import OrderedDict
+from odict import OrderedDict
 
 # Reference: https://www.kunxi.org/2014/05/lru-cache-in-python
 class LRUDict(object):
--- a/core/enums.py
+++ b/core/enums.py
@@ -7,7 +7,7 @@ See the file 'LICENSE' for copying permission
 
 import sys
 
-from thirdparty import six
+import six
 
 class _(type):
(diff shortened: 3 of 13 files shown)


stamparm:
Thank you for your work, though, you've oversimplified this one:

A) `core` is not the Maltrail. `core` is only holding the core functions required by the both `sensors` and `server`
B) `thirdparty` is usable for (e.g.) users using `git` method (currently all of users). I don't see a problem of it staying there as is because those same libraries are not versatile in updates nor features
C) There is a bigger issue here, which is not related to the PyPI structuring and that is the installation of Maltrail as a service on generic *nix systems. Just `pip install maltrail` won't do much in current state

jayvdb:
The intent isnt for this to be final, nor for `pip install maltrail` to become a solution for installation of `maltrail`.  Both are possible, but would be more disruptive to the current codebase, and would be a wasted effort to do unless you are interested in publishing to PyPI, and providing support for installation from PyPI, which may not be appropriate at the moment.

The objective is solely to make some of the maltrail functionality available to other applications, as described at https://github.com/stamparm/maltrail/issues/6297 .  I would use this immediately, and I suspect others would also, and with more app developers relying on maltrail, hopefully that results in more people helping with maintaining it also.

If you do want sensors and server included in the PyPI package, I can do that also.  Then the PyPI sdist would be approaching what distro packagers could be able to consume, and of course they would add the service files needed.

Including `thirdparty` in a PyPI package is possible too, and such vendoring is not uncommon, especially of six.  I can do that also, and in my distro packaging I then just patch to the source to de-vendor these.

jayvdb:
@stamparm , could you please give me more direction on how to proceed.  I would really like to use maltrail's features as part of a library I am building without resorting to copying&pasting them.

stamparm:
@jayvdb as far as I can see, you are interested ONLY into trails, because bundling ONLY the core doesn't make any sense. There is no point in making `maltrail` PyPI package which will hold only trails inside (along with dummy core). Hence, I'll need to drop this until I do it properly (for all the functionality)

jayvdb:
I am also using a few parts of core, which is why I included it.

As indicated above, I am happy to do a more complete version if that is what you want, and your last post indicates that, so I will do that today.

stamparm:
@jayvdb please, I'll do this when I catch time. I just need to think about it to do it properly

PR #6350, opened 2020-01-24
Create pupyrat.txt

MikhailKasimov:


stamparm:
Last part looks suspicious. Compromised sites?

MikhailKasimov:
Minus ```caltech.fr``` and ```stephanemalka.com``` -- they are indeed legit.

PR #6355, opened 2020-01-24
Update emotet.txt

MikhailKasimov:


stamparm:
All these are already inside

PR #6392, opened 2020-01-27
Create bdf_injector.txt

MikhailKasimov:
Do not want to put it to ```generic```. Name is based on Kaspersky's detection (```Packed.Win32.BDF.a```) and behaviour of the packer itself (```Malware-Cryptor.Inject.gen```).


stamparm:
A) Packer is not malware. Also, different malware can use this single packer. There are hundreds of different packers, obfuscators,... It doesn't make any sense to treat them the same as malware.
B) This is generic trail as it gets

MikhailKasimov:
OK, superseded by #6393

PR #6412, opened 2020-01-27
Create woof.txt

MikhailKasimov:


stamparm:
-> `malicious/woof.txt` (similar to malvertising and crypto jacking)

MikhailKasimov:
Resounded in #6421

PR #6495, opened 2020-01-30
Update apt_tvrms.txt

MikhailKasimov:
"TeamViewer"-based attacks can live in this trail, I suppose. Will separate, if new trail is preferable here.


MikhailKasimov:
Will go to ```rmsrat```

PR #6504, opened 2020-01-30
Merging Ursnif, Gozi and Dreambot

MikhailKasimov:
Merging Ursnif, Gozi and Dreambot into one single trail, because of being the same mal-family.


stamparm:
`DreamBot is a malware that extends the functions of Ursnif (Alias: Gozi) and it is mainly targeting financial institutions to steal authentication` - https://www.lac.co.jp/english/report/2017/10/16_alert_01.html

stamparm:
I am not pro merging of variants which are after a while creating a totally different malware types

stamparm:
Single reason why this could make sense is because people and AV tend to mix those in "findings". Seen a lot of those cases

PR #6552, opened 2020-02-02
Issue 2644

MikhailKasimov:
Inheriting SQL-detection regex from ```settings.py```. Hope, this will work for UA. Ref: https://medium.com/@frostnull/sql-injection-through-user-agent-44a1150f6888

### DIFF
--- a/misc/ua.txt
+++ b/misc/ua.txt
@@ -1247,3 +1247,8 @@ IEMoblie
 # Reference: https://twitter.com/reecdeep/status/1218098821143703552
 
 KSKJJGJ
+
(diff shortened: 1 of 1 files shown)


stamparm:
I believe that we had a similar discussion. SQL injection attempts in UA are last attempt(s) from attackers. Also, the probability of success is miniscule. There is no need for something spectacular here. I'll put a trivial version of this

MikhailKasimov:
Yes, we had. This PR is just a little bit "childish" attempt to bump the topic with some implementation of it.

```I'll put a trivial version of this``` <-- Thnx!

PR #6575, opened 2020-02-03
Update agenttesla.txt

MikhailKasimov:


stamparm:
Please, be careful with agenttesla IOCs as those are in almost all cases just compromised domains. Also, I've opened 3 random domains from here and all were old and compromised

MikhailKasimov:
Ok, will re-check domains a little bit later.

MikhailKasimov:
Have cleaned the list.

stamparm:
Sorry, have to drop this one down. Tried again 3 random domains, all older than couple of years. I've said my opinion of AgentTesla in previous comment

PR #6614, opened 2020-02-05
Update settings.py

MikhailKasimov:
Tracking potential password disclosure via open HTTP, e.g. in POST-requests. Inspired by https://twitter.com/bad_packets/status/1222784829332131840/photo/1 , https://i.imgur.com/qf5k6FF.png and similar functionality in AlienVault.


stamparm:
Problem is that this will extract passwords for even regular user attempts. Something that I am not keen of (even in SIEM solutions)

MikhailKasimov:
```Problem is that this will extract passwords for even regular user attempts.``` <-- Well, generally yes. But currently browsers alarm when user tries to authenticate via HTTP, SIEMs mark such cases as insecure. Various of security standards tell not to put plain-text passwords via insecure protocols... So, personally, I cannot see that problem is big here. Such alarm can be useful for cases of some trojan stealers. Let's have a test on this functionality here.

stamparm:
A) `SIEMs mark such cases as insecure` <- they mark them as potential `POLICY` issue
B) `I cannot see that problem is big here` <- I don't like to see this kind of stuff in SIEMs
C) `Such alarm can be useful for cases of some trojan stealers` <- plenty of user entries (hence, FPs)

PR #6615, opened 2020-02-05
Update lokibot.txt

MikhailKasimov:


stamparm:
Please use full paths if suspected of compromised sites (Tip: if there are >5 URLs, then most probably you'll have compromised sites in there)

PR #6651, opened 2020-02-08
Create clipbanker.txt

MikhailKasimov:


MikhailKasimov:
```The mining pool for the miner, with port manip2[.]hk:7777.``` -> moving to ```crypto_mining.txt```

PR #6663, opened 2020-02-08
Update crypto_mining.txt

MikhailKasimov:


stamparm:
This is more a `malware` stuff than `suspicious`. Also, `crypto_mining.txt` should contain only general crypto-mining pools, while this seems to be a case-specific

MikhailKasimov:
OK, where to move?

MikhailKasimov:
Closed via #6690

stamparm:
As per https://www.cybereason.com/hubfs/BitBucket%20Malware%20Arsenal%20IOCs.pdf, this is a mish-mesh of different malware. Also, those VT references don't have a specific (non-generic) name. Maybe just to put it into the `generic.txt`?

MikhailKasimov:
I initially sorted this ```bucket``` to keep ```azorult``` and others to their own trails. This one is unknown and, while port 7777 is specific, I decided to put Kaspersky's|NOD32's detection name. Let it live in dedicated trail. If somewhat -- will move or merge with others.

PR #6669, opened 2020-02-08
Create install_capital.txt

MikhailKasimov:
```# Reference: https://twitter.com/benkow_/status/1225786810330222594```
```# Reference: https://twitter.com/benkow_/status/1225784901502472192 (Reason of putting this as dedicated trail instead of ```pua.txt```: ```When AdWare Becomes Pay-per-Install Cyber-Crime.``` (c))```
```# Reference: https://pastebin.com/LhVBum2q (``` C&C domains from 2017 to 2020```).```


stamparm:
1) `The security industry has been indulgent with PPI for years considering it just as adware-related but the reality is very different, these networks are potentially huge malware distributors frequently used by various cyber-criminals.` -> `malicious/install_capital.txt`
2) `We observed more than 500 offers and almost 200,000 domains between 2017 and 2020 and all those domains point to the same IP: 54.88.21[.]193.` -> maybe it would make sense to just compress everything to a single IP (`54.88.21.193`) as your initial commit contains almost 200 thousands domains (not sure that its worth to have them all included - taking too much users memory for a something that is not worth of it)

MikhailKasimov:
Resounded in #6691

PR #6735, opened 2020-02-12
Update lokibot.txt

MikhailKasimov:
Two sites are legit compromised.


stamparm:
p.s. we had those conversations. If particular directory is being used (e.g. `http://46.21.147.207/primeight/logs/fre.php`) that's a more valuable trail than the IP itself as that same IP will be later reused for legitimate purposes

MikhailKasimov:
> /fre.php

This generic will cover this and it lives hundrends of years for Lokibot. We spoke on ```/sab.php``` and this trail claims to have some /subdirectory/. That's why I use pure IP -- potentially MT could cover other malwares related to this IP.

PR #6736, opened 2020-02-12
Update generic.txt

MikhailKasimov:
This live hundreds of years in ```generic```, ```email_stealer```, ```ta505```, ```amadey```.


stamparm:
There is no more `email_stealer`. It is now merged into `generic` (we spoke over this)

stamparm:
Reopening because of `ta505` and `amadey`

PR #6744, opened 2020-02-13
Update lokibot.txt

MikhailKasimov:


stamparm:
Please don't use this as there are numerous cases of compromise (particularly in case of trails with URLs). Also, just a note, comments start with `#`

PR #6769, opened 2020-02-14
Update bad_service.txt

MikhailKasimov:
```107.189.7.176``` returns ```crypters.info```, but if to put ```http://107.189.7.176/crypter/``` -- it returns subfolders, which can contain POST-ed info from backdoors. Simultaneously ```crypters.info``` is not malicious by itself. So -- ```bad_service.txt```.


MikhailKasimov:
UPD: Domain will be detected as ```bad_service.txt```, but IP -- as ```generic.txt``` one. See https://github.com/stamparm/maltrail/pull/6759/commits/b864970b1ac722272fc51de8a1c95dc295f75df5

PR #6802, opened 2020-02-16
Update pua.txt

MikhailKasimov:
OSX Adware


MikhailKasimov:
Resounded in https://github.com/stamparm/maltrail/pull/6803

PR #6830, opened 2020-02-17
Update lokibot.txt

MikhailKasimov:


stamparm:
p.s. if there are more than 20 IOCs, be sure that there are lots of compromised servers inside (e.g. assemba.co.uk, brodam.ro, centrehotel.vn, etc.)

MikhailKasimov:
OK, because I've cleaned some compromised sites. Thanks!

PR #6848, opened 2020-02-18
Update ursnif.txt

MikhailKasimov:
Despite on being declared as ```ransomware``` (without name) and having tail ```/pagkit56.php```, which  is very similar to ```ursnif```-based variations, let it all live in ```ursnif.txt```.

### DIFF
--- a/trails/static/malware/ursnif.txt
+++ b/trails/static/malware/ursnif.txt
@@ -2233,10 +2233,6 @@ lapenik.at
 g4xp7aanksu6qgci.onion
 l35sr5h5jl7xrh2q.onion
 
-# Reference: https://twitter.com/reecdeep/status/1229674592068866048
-
-mengather.com
-
 # Generic trails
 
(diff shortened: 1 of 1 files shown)


stamparm:
A) https://app.any.run/tasks/362e833e-6eb1-47f7-8f4c-40949a02c9dc/ - modified files 2353
B) -> `generic.txt` (p.s. it uses extension ROGER, hence you can use `roger.txt`)

MikhailKasimov:
OK, will move domain there. Let tail live here.

stamparm:
Have to close this. Tail can't live in unrelated malware place

MikhailKasimov:
Why ```not related``` if it is similar (heur) to ```ursnif``` ?!

stamparm:
Mixing of two malware families because of "similar" trail(s) doesn't seem right. Somebody will search for that same trail in google and (s)he will be presented with ursnif name, while totally unrelated

PR #6929, opened 2020-02-21
Update nivdort.txt

MikhailKasimov:


stamparm:
We've already had similar discussions. Putting every DGA combination into any kind of network traffic monitoring system (even in Maltrail) is not considered as "efficient". Using tons of user's memory, with little to no benefits

PR #6951, opened 2020-02-22
Update tofsee.txt

MikhailKasimov:


stamparm:
Please use `#` for start of comment. I can see that you are using comment inside the comment, but just to have to concise across everywhere

PR #6965, opened 2020-02-23
Update predatory.txt

MikhailKasimov:
Some optimization.


stamparm:
Introducing too many IP based trails from a bulk export could introduce lots of FPs in the long run

PR #7068, opened 2020-02-27
Update netsupport.txt

MikhailKasimov:


stamparm:
If >10 IOCs in a report, **especially** if with full-paths, 99.99% be sure that majority are **compromised**

PR #7145, opened 2020-03-02
Update pyrogenic.txt

MikhailKasimov:


stamparm:
In that same referenced app.any.run you have a real C&C `congresoambientalcanacintra.com`

PR #7173, opened 2020-03-03
Update generic.txt

MikhailKasimov:
```hokage.ru``` is already in ```azorult```, ```sakari``` and ```sorano``` stealers trails.


stamparm:
I removed `hokage.ru` from others and put it into `generic.txt` (yesterday)

PR #7175, opened 2020-03-03
Update whitelist.txt

MikhailKasimov:


MikhailKasimov:
Won't meet in practice:
```
nslookup panwdemo.duckdns.org
Server:  router.asus.com
Address:  192.168.31.1

Non-authoritative answer:
Name:    panwdemo.duckdns.org
Address:  192.168.2.187
```

PR #7222, opened 2020-03-05
Remove FP secure-magenta1.be2bill.com from magentocore.txt

monotypical:
This domain is owned by the payment processor Dalenys Payment and has been in use since 2011 http://whois.domaintools.com/be2bill.com

From the source (https://twitter.com/jknsCo/status/1200371937324740609/photo/3), the skimmer is served from `sixthjune.com`, submits credentials to `103.139.113.34` then redirects to `secure-magenta1.be2bill.com` which is the real payment gateway

### DIFF
--- a/trails/static/malicious/magentocore.txt
+++ b/trails/static/malicious/magentocore.txt
@@ -775,10 +775,6 @@ theresevit.com
 
 googlemgrteg.com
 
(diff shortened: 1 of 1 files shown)


stamparm:
Done. Sorry for the incovenience

MikhailKasimov:
@monotypical Thanks for pointing this out and sorry for FP!

PR #7268, opened 2020-03-07
Create phish_dns.txt

MikhailKasimov:
Trail for DNS-servers list, which are in use of various phish attacks, but without DNS hijacking attacks.


stamparm:
1) I am not convinced here. These same DNS servers can be used for valid stuff.
2) Example, `5.132.191.104` is an OpenNIC Public Server (https://servers.opennicproject.org/)

MikhailKasimov:
Minus ```5.132.191.104```

PR #7271, opened 2020-03-07
Update nemty.txt

MikhailKasimov:


stamparm:
1) Majority of these are FPs (e.g. `namebright.com` is a big provider)
2) We've already had couple of similar situations where I stated for N times that Talos intelligence is not doing a great job in this area

PR #7275, opened 2020-03-07
Update whitelist.txt

MikhailKasimov:
Removing this from whitelist, because this is ```elf_mirai``` (lives here) in fact: see ```URL``` sections in https://www.virustotal.com/gui/domain/scan.casualaffinity.net/relations and https://www.virustotal.com/gui/domain/jhasdjahsdjasfkdaskdfasbot.niggacumyafacenet.xyz/relations

### DIFF
--- a/misc/whitelist.txt
+++ b/misc/whitelist.txt
@@ -2781,11 +2781,6 @@ reseau-js.com
 coloros.com
 sagawa-exp.co.jp
 
-# Exceptions for abuse.ch (dirty patch for addresses used by outside scanners introducing noise being recognized as 'malware')
-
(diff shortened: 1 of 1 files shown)


stamparm:
There are problematic. I've left the comment in that commit. Outside scanners are using it for scanning the Internet. For example:

rrr

These requests can't be marked as malware. They are just noise

MikhailKasimov:
OK, will also clean from ```elf_mirai.txt``` trail.

MikhailKasimov:
https://github.com/stamparm/maltrail/pull/7291

PR #7417, opened 2020-03-13
Update ta505.txt

MikhailKasimov:


stamparm:
Not TA505 related (https://github.com/SherifEldeeb/TinyMet/blob/master/Source.cpp)

PR #7429, opened 2020-03-14
Update raccoon.txt

MikhailKasimov:


stamparm:
If `http` then it is better to use full paths (already had similar discussions). Also, in these cases, it is really easy to extract paths

stamparm:
p.s. these same HTTP servers are all down, except the last one. Thus, it would be better to just put generic trails and/or generic trails with IP/paths. As said in many occasions before, sole IP is a worthless indicator. Anything more (e.g. port, path, domain...) is far more valuable and "enriches" the IP/context

PR #7488, opened 2020-03-17
Update settings.py

MikhailKasimov:
Based on: https://www.virustotal.com/gui/file/f02de53011911ce236fd4aa12356da4a47e0632cedd48bd150d2b99ed79890c8/behavior/Tencent%20HABO


stamparm:
Not for this as it would generate noise (without too much effect)

MikhailKasimov:
OK, declined.

PR #7494, opened 2020-03-18
Update c2_panel.txt

MikhailKasimov:
```inurl:"/IRemotePanel"``` says OK to get added as detection sign.


MikhailKasimov:
Moved to #7495

PR #7797, opened 2020-03-27
Create sinkhole_reversinglabs.txt

MikhailKasimov:
```Currently it redirects to a sinkhole server with the IP address 172.105.123.10.``` -- from https://blog.reversinglabs.com/blog/unpacking-kwampirs-rat

Related to #7796

### DIFF
--- a/trails/static/malware/sinkhole_reversinglabs.txt
+++ b/trails/static/malware/sinkhole_reversinglabs.txt
@@ -1,0 +1,7 @@
+# Copyright (c) 2014-2020 Maltrail developers (https://github.com/stamparm/maltrail/)
+# See the file 'LICENSE' for copying permission
+
+# Reference: https://blog.reversinglabs.com/blog/unpacking-kwampirs-rat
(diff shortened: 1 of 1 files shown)


stamparm:
A) I am not sure that this is Reversing Labs sinkhole.
B) There is a funny domain `siar.us` at https://www.virustotal.com/gui/ip-address/172.105.123.10/relations, where its subdomain `r7.siar.us` points to the sinkhole, while the (e.g.) `a1.siar.us` doesn't. This basically means that the owner of `siar.us` (which is not Reversing Labs) is pointing that same domain to that same IP

stamparm:
By Googling I can see that other companies pin-pointed to KwampirsRAT at start of this year. Thus, I really believe that ReversingLabs are not "first" looking into this

stamparm:
-> `sinkhole_unknown.txt`

MikhailKasimov:
Superseded by #7847

stamparm:
p.s. Just found that `SIDN Labs` also sinkhole 3 of those domains (176.58.104.168). Thus, proving that other entities are interested into this RAT too

PR #7871, opened 2020-03-31
Update parasitehttprat.txt

MikhailKasimov:


stamparm:
Referenced link doesn't mention parasitehttprat. Also, it is a mish-mesh of different malware types

MikhailKasimov:
> Referenced link doesn't mention parasitehttprat.

See also string 32 : ```# Reference: https://github.com/silence-is-best/c2db#parasite-stealer-aka-nexus``` . Will update ```# Aliases:``` field too.

stamparm:
Okie dokie. Seen a minute ago

PR #7889, opened 2020-04-01
Update domain.txt

MikhailKasimov:
Many ```apt_kimsuky``` C2s are hosted on ```mireene.com``` -- somesubdomain.```mireene.com```. Not sure for ```mireene.com``` being malicious itself, but connections to this domain should be considered as suspicious.


stamparm:
Just by Googling and visiting: https://www.virustotal.com/gui/domain/mireene.com/relations

A) 250,000 results in Google search
B) Top100k domains
C) Only 1 communicating malware sample, which fetches `http://hiperv.mireene.com/simple_hotkey_list_190710.html` (`hiperv.mireene.com` actually is related to malware)

PR #7968, opened 2020-04-03
Update emotet.txt

MikhailKasimov:


stamparm:
I don't see malware/sample which communicates with those ports. Also, dates are way in past. Will "modify"

PR #7998, opened 2020-04-04
Update qakbot.txt

MikhailKasimov:


stamparm:
This doesn't look good. It seems like a decoy traffic to me to hide a real C&C

stamparm:
One of related samples (https://www.virustotal.com/gui/file/d6cda4b65efa7136a6c269f681196aa63935c93d959dbb9bdf338ba6ed69b5d6/behavior/QiAnXin%20RedDrip) is called `TheUntraceable RAT.exe`. Anyway, author(s) is trying to mask the real C&C(s) behind all this junk traffic

MikhailKasimov:
OK, can be closed. Will add ```ip-adress.com``` to ```ipinfo.txt``` trail -- we do not have this one.

stamparm:
Anyway, as I see that sample(s) are not doing a random generation of those IP:port combinations, I would reduce the "junk"/"noise" to contain only the non-80 and non-443 combinations. That way it would be reduced to something meaningful

stamparm:
`ip-adress.com` is an interesting thing that I've also learned from this example :). It seems that author(s) are trying to do an "interesting" evasion with this one. I could do some generic stuff, like we do for the `onion` trails

MikhailKasimov:
> Anyway, as I see that sample(s) are not doing a random generation of those IP:port combinations, I would reduce the "junk"/"noise" to contain only the non-80 and non-443 combinations. That way it would be reduced to something meaningful

OK, I currently do not delete this branch.

MikhailKasimov:
> `ip-adress.com` is an interesting thing that I've also learned from this example :). It seems that author(s) are trying to do an "interesting" evasion with this one. I could do some generic stuff, like we do for the `onion` trails

#8018 (ipinfo)

```I could do some generic stuff, like we do for the `onion` trails``` -->OK. Thnx!

MikhailKasimov:
Just to be linked: https://github.com/stamparm/maltrail/commit/0e4736648c1876c3fd0ce6a28d19c98c9b354715

PR #8023, opened 2020-04-06
Update generic.txt

MikhailKasimov:


stamparm:
1) Scanner (as pointed in twitter comment)
2) I can't find anything at port 80 (at least currently)
3) VT doesn't show any signs of communicating samples. Also, those domains in passive records are more than couple of months old (related to previous owner)

PR #8141, opened 2020-04-12
Update crypto_mining.txt

MikhailKasimov:


stamparm:
`140.82.57.249`== `googdns.ml` (https://www.virustotal.com/gui/ip-address/140.82.57.249/relations) - pure malware related (seems like Remcos RAT based on `svchost.club`

PR #8151, opened 2020-04-13
Update kraken.txt

MikhailKasimov:
After taking a look on ```URL``` section of https://www.virustotal.com/gui/domain/blasze.tk/relations , entire domain should be under detection. Tails ```/....``` are too various. Will catch them all now.


stamparm:
This URL is specific for `kraken`. Domain is not. Domain is generally used for IP logging, not malware

PR #8215, opened 2020-04-15
Update apt_patchwork.txt

MikhailKasimov:


stamparm:
p.s. too many IPs, while it is clear from sample(s) that domain are being used to access C&Cs

PR #8271, opened 2020-04-18
Update cerber.txt

MikhailKasimov:
Merging TeslaCrypt regex with Cerber's one.


stamparm:
Reason to merge it into `cerber`?

MikhailKasimov:
Reason is the unification of detection regexes have one single regex instead of two similar ones (just compare ```cerber``` and ```teslacrypt``` onles): 

```some-set-of-symbols.subdomain.domain```

Screenshot_128

Hence, from one side, MT will cover both of ```cerber``` and ```teslacrypt``` by ```some-set-of-symbols``` in one single regex **with memory usage decrease** and, from another side, MT will cover ```teslacrypt``` by its ```.at``` domains (#8270) and their ```some-set-of-symbols```, which are currently could be absent in detection regex.

stamparm:
1) The memory is not a factor here, as regex is regex, no matter where you put it. Also, all the regexes are merged at the end (Note: I only nagged that we are limited to 100 regexes because of Python's re mechanism)
2) To me, something is fishy here. This "same set of symbols" is too UNIQUE. Thus, either somebody made a mistake here (I can see that you are again referencing Talos, which I have zero confidence) or one of malware "gangs" decided to troll the other one. By the rule of Occam's razor, I would go straight to the simplest solution - somebody did a mistake.
3) I hope that you'll realize at some point that Talos has maybe one or two extraordinary guys, which are most probably external contractors, while the rest of the whole ("malware research") lab is a pure **shit**. I stand by this words because they proved this to me on numerous occasions

stamparm:
1) Now going through https://github.com/stamparm/maltrail/commit/2127f23b8506a3cfa51b9700443154dcf3d11ad2#diff-3e18258b27f84b97c91bc2c9e744f74f and I can see only TeslaCrypt domains inside
2) Thus, while at the end everything is merged into one big regex, I do not see any valid reason to put that regex and merge into `cerber.txt`.
3) When taken 1) and 2) into the consideration, the only result of all this will be the wrong recognition of those TeslaCrypt domains as Cerber. Regex is indeed merged and run as ONE, but Maltrail is preserving the malware name based on the regex run

MikhailKasimov:
Talos is put of scope here, because everything was more simple: I've noticed to myself, that ```cerber``` and ```teslacrypt``` regexes are very similar to each other by reason of domains structure: ```some-bla-bla-bla.subdomain.domain```, and generally could be megred in one single regex. In case of ```This "same set of symbols" is too UNIQUE``` I revert my deletion of TeslaCrypt's regex in #8304 and current pull request can be closed.

```p.s. (new) Afterwards I've seen that you've put a Talos reference for a totally different domain, but I get easily triggered by even a mention of that "group of people"```

I have rather cold blood on Talos, because I use it as every other source, that could contain useful information for malware detection. Of course, minus real idiotism like marking ```169.254.x.x``` IP as malicious and so on.

PR #8372, opened 2020-04-23
Installation guide for different Linux distros

MikhailKasimov:
Installation description update for different Linux distros (Debian/Ubuntu and SUSE/openSUSE).

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,8 +263,20 @@ Note: **Server** component can be skipped altogether, and just use the standalon
 
 The following set of commands should get your Maltrail **Sensor** up and running (out of the box with default settings and monitoring interface "any"):
 
+- For **Debian/Ubuntu**
+
 ```
 sudo apt-get install git python-pcapy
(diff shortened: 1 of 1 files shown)


stamparm:
```
sudo zypper install gcc gcc-c++ git libpcap-devel python2-pip python-devel python3-devel
sudo pip2 install pcapy
git clone https://github.com/stamparm/maltrail.git
cd maltrail
sudo python sensor.py
```

1) Why installing both python2 and python3 devel packages?
2) There is no need for explicit Python2 environment, thus, if there is a way to just choose default python packages (like in case of Debian/Ubuntu), that would be the best

MikhailKasimov:
Because I have\use openSUSE, and these packages were needed to have normally working MT sensor.

stamparm:
Both Python2 and Python3 `devel` packages?

MikhailKasimov:
Yes.

PR #8397, opened 2020-04-24
Update quasarrat.txt

MikhailKasimov:


stamparm:
I would not mind if VoidRAT goes to the separate file

PR #8399, opened 2020-04-24
Update gh0strat.txt

MikhailKasimov:


stamparm:
1) All domains are older than couple of years
2) Last 2 have tens of thousands of backlinks on Google
3) `talosintelligence` - not sure how many times will we repeat this whole story again

PR #8410, opened 2020-04-25
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
1) I've already removed couple of similar "dynamic domains". This one will trigger lots of hits, while providing back little to no useful hits.
2) Referenced case is an isolated case
3) This is tied to a general service provided by QNAP and not a general "dynamic domain" provider

PR #8440, opened 2020-04-27
Create python_generic.txt

MikhailKasimov:


stamparm:
1) We already had this file. The final "conclusion" was that it doesn't make any sense (any programming language, especially interpreter one, would require a separate file)
2) -> `generic.txt`

MikhailKasimov:
In ```generic``` I will forget to replace, if\when explicit name is given. In case of ```python_generic``` I would merely change ```_%{name}%``` part. It is just as a reminder for myself.

MikhailKasimov:
Fixed in https://github.com/stamparm/maltrail/pull/8440/commits/064919db638aed1e88413255e77cfaa6f755af32

MikhailKasimov:
Replaced via https://github.com/stamparm/maltrail/pull/8441

PR #8442, opened 2020-04-27
Update zloader.txt

MikhailKasimov:


stamparm:
DGA, thus, those domains are only valid for a specific date

PR #8499, opened 2020-04-30
Update netsupport.txt

MikhailKasimov:


stamparm:
p.s. I am more and more prone to an idea of using `http://` which you are deleting here. Reason can be found in your commits where in the same time you refer access to HTTP servers via IP with `http://` prefix (e.g. `http://123.123.123.123`), while removing that same for cases when accessing HTTP servers via domain (e.g. `foobar.com/path`)

MikhailKasimov:
My thought is that domains anyway accessable, so using ```http://``` or ```https://``` for them is redundant. Especially ```http://www.```: MT will detect ```www.domain``` and will miss ```domain```. Otherwise it wll catch both of variants. For IPs -- using ```http://``` is our agreement somewhere around in 2017, when we hadn't IP:Port scheme in release. I'm keeping this agreement.

PR #8501, opened 2020-04-30
Update ursnif.txt

MikhailKasimov:


MikhailKasimov:
Trickbot instead: #8502

PR #8505, opened 2020-04-30
Update fareit.txt

MikhailKasimov:
Also cleaning redundancy: this generic detection sign lives in ```fareit.txt`` for hundred of years. :)


stamparm:
We discussed this for many times. In case of sites referenced with only an IP address it is far safer to reference the whole URL (with path) because that same IP will be used in future by somebody else

MikhailKasimov:
Fixed in https://github.com/stamparm/maltrail/pull/8505/commits/22d105a79cbab6833efb2e66a54b89e2937a2cdf

PR #8546, opened 2020-05-02
Create socelars.txt

MikhailKasimov:


stamparm:
1) https://blog.malwarebytes.com/detections/spyware-socelars/ - `Spyware.Socelars is Malwarebytes’ detection name for a large family of malware that is capable of stealing sensitive information from Windows systems.` This all seems too generic
2) There is nothing specific regarding "Socelars" on Google. No usable reports

MikhailKasimov:
If ```stealer.txt``` is present, it will be more easy to detect. OK, will move to ```generic.txt```

MikhailKasimov:
Resounded in https://github.com/stamparm/maltrail/pull/8550

PR #8549, opened 2020-05-02
Create schwarzesonne.txt

MikhailKasimov:


stamparm:
Inside of community comments: `Testing people's detection of pandora rat. forgot to login first. This is a test sample only. `

MikhailKasimov:
Couldn't get -- don't we need to detect this?

PR #8655, opened 2020-05-05
Update for data leak heur on Android.Fanta

MikhailKasimov:
From ```https://www.group-ib.ru/blog/fanta``` (translation from Russian):

```divice_admin – поле определяет, были ли получены права администратора. Если права администратора были получены, поле равно 1, иначе 0```

 -->

```divice_admin – field value defines whenever admin rights are on|off on device. In yes, divice_admin=1, otherwise=0```.

Plus updating telephone number and company\operator parameters: ```mode=register_bot&prefix=2&version_sdk=<%VERSION_SDK%>&imei=<%IMEI%>&country=<%COUNTRY_ISO%>&number=<%TEL_NUMBER%>&operator=<%OPERATOR_NAME%>```.

Related to https://github.com/stamparm/maltrail/pull/8651/commits/d20ed92544fffa1827edd0c89d6b498dde56b02d


stamparm:
1) Please, do not over-extend this data-leakage heuristics as we could end with tons of noise in reports. All telemetry, add-companies, fishy javascript for user platform detection,... are leaking info. Please, KISS
2) `number` and `operator` would results with tons of FPs, while `divice_admin` is case specific (i.e. there are tons of similar parameters in similar cases where the other side is trying to find as much info on user as it can)

PR #8671, opened 2020-05-06
Update zloader.txt

MikhailKasimov:
After the second look -- let this also be detected: this will mean, we hadn't miss report of known infection case.


stamparm:
1 day of DGA is worthless as that time passed long time ago

PR #8720, opened 2020-05-07
Update settings.py

MikhailKasimov:
See ```&cmd=wget ``` and ```/shell?wget``` in https://twitter.com/bad_packets/status/1258503413836312576 with some generalization for popular upload/download tools . Hope. it is applicable and not "over-extending". Please, fix regex, if needed.


stamparm:
`wget` is already covered by current regex

PR #8730, opened 2020-05-08
Update virut.txt

MikhailKasimov:


stamparm:
1) Virut DGA
2) We already use couple of DGA feeds (e.g. Bambenek and 360) which get fresh lists of DGA domains

MikhailKasimov:
Weird, but all previous DGA-based PR-s for ```virut``` got merged, but not today. Why?

PR #8744, opened 2020-05-08
Update dridex.txt

MikhailKasimov:


stamparm:
1) Dridex DGA (only valid for one day in past)
2) Talos Intelligence

PR #8752, opened 2020-05-09
Update android_sandrorat.txt

MikhailKasimov:


stamparm:
1) Sandrorat is a variant of Androrat. Not sure that we should merge all variants of malware together (I know that we did couple of merges), because newer variants have different set of features than the older/base ones
2) There is already `android_androrat`

MikhailKasimov:
I can't get -- how all of this blocks to add trail into ```sandrorat``` instead of ```androrat```?

MikhailKasimov:
Thnx!

PR #8867, opened 2020-05-14
Update qakbot.txt

MikhailKasimov:
All IP connections go to dedicated section


stamparm:
1) We are already "used" to this kind of intermix between domains and IP connections (e.g. Emotet)
2) This same approach/change looses any kind of reference information. Just a bunch of IP:ports out of nowhere

MikhailKasimov:
I've just collected all ```IP:port``` we already merged in one single section with deduplication. That's all I've done.

stamparm:
In this form `Reference`s are missing. Someone looking into will just see a bunch of IP:port entries, no matter that you know that they are related to references from up above

stamparm:
p.s. same thing applies to Emotet. I strongly prefer the current form where the current reference shows the source of the information. Bunch of IP connections, with lost information where those came from, is inconsistent to all the other trails in all other files (i.e. Reference is a source of information)

MikhailKasimov:
I propose next:

1) Current PR gonna be closed.
2) I do new PR with last trails to add (with IP:port).
3) You help me to clean IP:port dups up, like we cooperate for Emotet.

Is such algo acceptable?

PR #8942, opened 2020-05-20
Update crypto_mining.txt

MikhailKasimov:


stamparm:
A) Malware specific, while `crypto_mining` should be for general crypto-mining stuff
B) Already covered by generic trails

MikhailKasimov:
ok

PR #9029, opened 2020-05-27
Update settings.py

MikhailKasimov:
To be able to detect ```/pupmips``` trails. See https://www.virustotal.com/gui/file/73a39b4fa389eedadd230adf1b96ba7e295c30edbacd9484fef6168a0b345e37/community


stamparm:
It can be put inside the `elf_` specific file. This is not a generic file extension like the others

MikhailKasimov:
OK

PR #9061, opened 2020-05-30
Update apt_turla.txt

MikhailKasimov:


stamparm:
`Expect some false positives as it appears these are compromised wordpress sites.` (https://www.epicturla.com/blog/sysinturla)

MikhailKasimov:
Gonna close it, because I can't see full paths to detect these ```compromised wordpress sites```.

PR #9215, opened 2020-06-08
Update c2_panel.txt

MikhailKasimov:
Not sure about ```magecart```, but all IPs are allive.


stamparm:
-> `magentocore.txt` (https://blog.malwarebytes.com/cybercrime/2019/07/no-mans-land-how-a-magecart-group-is-running-a-web-skimming-operation-from-a-war-zone/)

MikhailKasimov:
Done in #9218

PR #9229, opened 2020-06-08
Update guloader.txt

MikhailKasimov:


stamparm:
A) It even has its own report
B) It even says in title `GuLoader? No, CloudEyE.`
C) -> `cloudeye.txt`

MikhailKasimov:
OK-OK-OK... #9250

PR #9325, opened 2020-06-12
Update ua.txt

MikhailKasimov:
1) ```The User-agent for some of the Nemty versions was the ‘Chrome’ string. The user-agent is hardcoded as a single string in the ransomware instead of using an original user-agent.```
2) ```The User-agent changed to a new one, “Naruto Uzumake”```


stamparm:
Mozilla 4.0 was already included. Produced tons of noise

PR #9347, opened 2020-06-13
Update whitelist.txt

MikhailKasimov:


stamparm:
Already covered with `172.253.0.0/16` (https://github.com/stamparm/maltrail/blob/master/misc/whitelist.txt#L1335)

PR #9357, opened 2020-06-13
Update maltrail.conf

MikhailKasimov:
To be able to catch cases like https://twitter.com/bad_packets/status/1271873601952571392 . See also #9356 .


stamparm:
Not enabling by default. Problem on huge networks

PR #9479, opened 2020-06-21
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
These are some cable/DSL IPs. I can't see that any of these "dynamic" domains is using different IPs

PR #9499, opened 2020-06-22
Update magentocore.txt

MikhailKasimov:


MikhailKasimov:
About ```paypalapiobjects.com``` --> https://www.virustotal.com/gui/ip-address/82.148.18.218/relations : the same schene of subdomains for other Magecart domains: ```google-anatytics.com``` , ```sucuritester.com```.

PR #9593, opened 2020-06-27
Update elf_mirai.txt

MikhailKasimov:


stamparm:
This doesn't look okish to me. Too much memory is being consumed with little to no benefit(s)

MikhailKasimov:
All previous the same generic commits were OK to get merged. What has changed today?

stamparm:
More than 2500 entries per `elf` pull request. That's what's changed. We already had this kind of problematic bulky pull requests before - and in majority of cases I disputed them. I don't see a great benefit here

MikhailKasimov:
You always merged all my such generic trails for ```Mirai``` and ```Gafgyt``` and they cover tons of IP/domains Mirai-based sources. I suspect if I "smudge" this 2500 by adding 50-100 in a day, you'd even do not pay attention on it...

```I don't see a great benefit here``` <-- if you can see better solution to detect such trails, please, welcome to put it to Maltrail code! I gonna be really glad. Seriously.

stamparm:
In case of `elf_mirai` there is already that regex in `settings.py`. In case of `elf_gafgyt` I don't see a point in covering every random port that some random malware operator is using per hourly/daily basis.

One thing. I am merging thousands of pull requests without any "issues" and now you trigger on my valid concern of adding memory clog/junk without any really benefit.

MikhailKasimov:
You've merged ~1200 the same ```junky``` records **by your own hands** in https://github.com/stamparm/maltrail/commit/a59ca300fddb002988cc3d65d61de64114ddf9c3 and you didn't complain on ```memory clog/junk without any really benefit``` and now -- wow! -- the similar PR gets ```memory clog/junk without any really benefit``` status. If you cand find more acceptable way to detect all this -- OK, this would be great.

PR #9611, opened 2020-06-27
Delete urlvir.py

MikhailKasimov:
Dead feed -- returns blank list via HTTP and HTTPS all together. Support doesn't give any info about their ```Export Database``` functionality.


stamparm:
Been aware. Thing is that I waited for them to return, but it seems that there is no return

PR #9614, opened 2020-06-28
Update elf_gafgyt.txt

MikhailKasimov:


stamparm:
What's the context? It seems to me that this kind of commits will go into user memory, but without benefit

MikhailKasimov:
Context is the same -- ```/Mozi.m``` and ```/.i```, which we cannot catch by ```CAPTURE_FILTER = ip or ip6```.

stamparm:
2345 new entries x 20 bytes (on average per entry) == 46900 bytes

I just see a problem here

MikhailKasimov:
If I add all this stuff is doses of tea-spoon (5-10 a day), they would be merged with no word against. :) But problem is A) we can't have ```ip or ip6``` (https://github.com/stamparm/maltrail/pull/9357), but we must detect such stuff in some way and B) previous similar PR (https://github.com/stamparm/maltrail/pull/9511/commits/a59ca300fddb002988cc3d65d61de64114ddf9c3) was merged with no words against. Personally I cannot see other solution on how to detect all this stuff.

PR #9733, opened 2020-07-07
Update purplefox.txt

MikhailKasimov:


stamparm:
AFAIK, there is EK purplefox and (fileless) malware purplefox. The same guys run the both. Thus, in your case (`casestudybuddy.club`) we are talking about the EK domain. Anyhow, I'll need to sort this out

PR #9771, opened 2020-07-09
Update elf_mirai.txt

MikhailKasimov:


stamparm:
I would put this into its own trails file (`moobot`). It has its own report. Also, `elf_mirai` is becoming really hefty. Somebody will think that Mirai is the single most active malware around, while actually it is the activity of its strains

MikhailKasimov:
OK, no problem at all.

MikhailKasimov:
Superseded by https://github.com/stamparm/maltrail/pull/9779

PR #9817, opened 2020-07-12
Create clidak.txt

MikhailKasimov:


stamparm:
-> `generic.txt`

MikhailKasimov:
I would like to avoid ```generic```, because of ```/~pete19c/r.php``` is being rather specific. If you insist moving to ```generic```, OK.

stamparm:
please into generic. `clidak` is a generic malware naming (it does not exists per se)

MikhailKasimov:
Done in #9842 and #9843

PR #9931, opened 2020-07-18
Create mailer.txt

MikhailKasimov:


stamparm:
Not malware, but the dropped hack scripts (i.e. SMTP mailing)

mailer

PR #9963, opened 2020-07-21
Create bad_ssh.txt

MikhailKasimov:


stamparm:
1) this is all related to the single malware-related incident
2) there are no "bad ssh" services here. only malware related backconnect SSH services
3) this should all be classified as `malware` with other name than `bad_ssh`. For example, https://www.reddit.com/r/Malware/comments/8bodwv/stumped_on_sample/ mentions "Gh0stRAT" because of the `Game over good luck by wind`

MikhailKasimov:
Replaced by https://github.com/stamparm/maltrail/pull/9971

PR #10063, opened 2020-07-26
Update settings.py

MikhailKasimov:
Case: 

```http://103.214.6.199/fuk4
http://103.214.6.199/fuk4t
http://103.214.6.199/fuk5
http://103.214.6.199/fuk6
```


stamparm:
We should really include generic stuff here, not some random names

MikhailKasimov:
Resounded in ```elf_mirai```:  https://github.com/stamparm/maltrail/pull/10064

PR #10160, opened 2020-08-01
Update android_ransomware.txt

MikhailKasimov:
Merging ```android_svpeng``` into ```android_ransomware```: it is locker by it behaviour, hence -- ransomware. ```koler``` and ```svpeng``` words gonna live in ```# Aliases:``` field.


stamparm:
I am not really for this idea. `android_ransomware` should be generic, and that's ok. Removing one non-generic malware `android_svpeng.txt` and mixing everything into `android_ransomware` doesn't look promising. Also, putting aliases for real malware names into generic trails file looks "unnatural"

PR #10269, opened 2020-08-08
Update crypto_mining.txt

MikhailKasimov:


stamparm:
both servers appear to be Electrum (AFAIK)

PR #10309, opened 2020-08-11
Update ua.txt

MikhailKasimov:
After the second look, gonna extend this regex to detect ```<iframe src=some/frame.html></iframe>``` too as a variant of loading some external code.


stamparm:
We had a discussion regarding this before. Please do not over-extend this as UA is rarely used in successful attacks which you are trying to cover here. `<script>` can stay but `<iframe>` is over-extending

PR #10314, opened 2020-08-12
Update ua.txt

MikhailKasimov:


stamparm:
will merge, but `fasthttp` looks like potential candidate for removal

MikhailKasimov:
OK, let's track it for some time

PR #10405, opened 2020-08-17
Update maltrail.conf

MikhailKasimov:
https://imgur.com/a/7gs10Af


stamparm:
fail2ban should result with blacklist for inbound attempts. also, the whole mechanism is tuned for extracting such IPs (not malware or dns changer or ....)

PR #10605, opened 2020-08-26
Update nivdort.txt

MikhailKasimov:


stamparm:
This is DGA. There is no point in committing thousands of domains that will be valid only for that day

MikhailKasimov:
OK, closing.

PR #10700, opened 2020-08-29
Create android_krbot.txt

MikhailKasimov:


stamparm:
krbot?

MikhailKasimov:
As I can see, there are Android samples only...

stamparm:
But I can't find any report on it. VT is not the best choice for picking a name

MikhailKasimov:
Yes, just https://www.microsoft.com/en-us/wdsi/threats/malware-encyclopedia-description?Name=TrojanSpy:AndroidOS/Krbot!rfn&ThreatID=2147760261 . But also these samples are also similar to ```android_spynote``` behaviour: connection to dyn-domain with specific port. In this case I would like to avoid giving ```android_generic```. So, let's decide what name to put here.

stamparm:
We are playing naming here. Generic is always the best place for this "new" and "unknown" samples. Later on they can be easily switched to another file once they come up in some other report

MikhailKasimov:
Ok, reverting. In https://github.com/stamparm/maltrail/pull/10709

PR #10725, opened 2020-08-31
Update mass_scanner.txt

MikhailKasimov:


stamparm:
Here should go regular (well-established) mass scanners. This subnet looks the same as some random chinese networks doing the regular scan all stuff

stamparm:
p.s. I have tons of similar cases like this. Just the regular Internet whitenoise

MikhailKasimov:
OK, closing...

PR #10788, opened 2020-09-02
Update necurs.txt

MikhailKasimov:
Google via ```inurl:"/forum/db.php"``` req shows OK to have such trail as generic.


stamparm:
Will merge, but Necurs uses DGA. That's why you get such different domains at each VT hit

PR #11015, opened 2020-09-12
Update settings.py

MikhailKasimov:
Reference: https://github.com/jas502n/CVE-2020-5902/


stamparm:
We had similar situations before. With this generic regex only generic attempts for code executions should be capture, not specific CVEs. Just to prevent over-extending without an actual need

MikhailKasimov:
OK

PR #11090, opened 2020-09-16
Update crypto_mining.txt

MikhailKasimov:
Part of https://github.com/stamparm/maltrail/pull/11089 (crypto-mining host)


stamparm:
1) That domain is already included in `mrb_miner`
2) `crypto_mining` is only for legit crypto mining services (for general usage, not just for single malware family)

MikhailKasimov:
> 1. That domain is already included in `mrb_miner`

No, it wasn't : https://github.com/stamparm/maltrail/pull/11099

PR #11123, opened 2020-09-26
Create kratos.txt

MikhailKasimov:


stamparm:
lucifer also uses `gate.php?hwid=`

MikhailKasimov:
Resounding in https://github.com/stamparm/maltrail/pull/11164

PR #11138, opened 2020-09-26
Update zeus.txt

MikhailKasimov:


stamparm:
first one is simda

MikhailKasimov:
> first one is simda

Done via https://github.com/stamparm/maltrail/pull/11163

PR #11408, opened 2020-10-07
Update domain.txt

MikhailKasimov:
Root form of domain from #11407


stamparm:
I don't see many malware sites there are VT reference. Maybe the better should be `free_web_hosting.txt`

MikhailKasimov:
> I don't see many malware sites there are VT reference. Maybe the better should be `free_web_hosting.txt`

Done in https://github.com/stamparm/maltrail/pull/11422

PR #11409, opened 2020-10-07
Update android_generic.txt

MikhailKasimov:
https://twitter.com/malwrhunterteam/status/1313798873270431744


stamparm:
Really too generic. If I see `/free.apk` in traffic I can't be confident that it is malware. But, if it will repeat in other incidents we'll put it here for sure

MikhailKasimov:
OK

PR #11494, opened 2020-10-11
Update elf_generic.txt

MikhailKasimov:


MikhailKasimov:
Resounding as dedicated trail: ```elf_powerghost.txt``` : https://github.com/stamparm/maltrail/pull/11497

PR #11609, opened 2020-10-16
Update emotet.txt

MikhailKasimov:
"Tails" are too various for ```Emotet```. IP-detection is more reliable.


stamparm:
Problem is that introduction of too many IP (or in this case HTTP://IP) specific trails could cause problems in near future (e.g. because of recycling of those IP addresses by some legitimate service)

PR #11654, opened 2020-10-18
Update whitelist.txt

MikhailKasimov:


stamparm:
Will put it into suspicious, just in case. Will remove later if FPs occur

MikhailKasimov:
OK

PR #11703, opened 2020-10-20
Update settings.py

MikhailKasimov:
https://www.virustotal.com/gui/file/038775e4cd93ef14bb696c6a09cd78c1131220fb934f084b3348b1c32e8b60d2/behavior/VirusTotal%20Jujubox


stamparm:
This is ancient. We can enable it and see if any FPs arise

MikhailKasimov:
OK, let's have a test here

PR #11704, opened 2020-10-20
Create lovesan.txt

MikhailKasimov:


stamparm:
1) This is Amazon EC2's IP. I am not sure what's the cause of contacting that IP. I can't find any info regarding it
2) This is ancient malware

PR #11888, opened 2020-10-28
Update js_script.txt

MikhailKasimov:


stamparm:
Not sure about these. It looks like regular pharma scam/spam

MikhailKasimov:
Generally. But, they compromise legit site to make redirection to their scam: https://twitter.com/unmaskparasites/status/1321506413995159552 That's the reason I put these domains into ```js_script```.

stamparm:
There are tons of similar redirect scams on compromised sites all over the web. Redirection itself is just a "blackhat seo" tactic, while JS does not have any intelligent logic inside

MikhailKasimov:
OK. Then I have cleaned trail: https://github.com/stamparm/maltrail/pull/11888/commits/0d93108cb56da9e3e64eecfee75ac07b62f47c9b

PR #11957, opened 2020-11-01
Create kozbot.txt

MikhailKasimov:


stamparm:
This doesn't look like a new strain/malware from the VT report

MikhailKasimov:
generic?

stamparm:
I would say that it is one of those usual MSIL RAT suspects, but as we don't have any information, the `generic` should be the most appropriate

MikhailKasimov:
Superseded by https://github.com/stamparm/maltrail/pull/11964/commits/19f1e170c13253af1a86e2f0d823a1e644f5c070

PR #12094, opened 2020-11-07
Update dynamic_domain.txt

MikhailKasimov:
Many of these domains simultaneously live in ```domain``` and ```dynamic_domain``` lists. Wanna do deduplication by moving them into one single place -- ```dynamic_domain```.


stamparm:
But lots of these are not "dynamic" as compared to our existing entries, even though the reference says: `http://trojanedbinaries.com/security/dynamicdns`

Dynamic domains should have its own provider gaining "dynamic DNS" service to anybody (registered)

stamparm:
I would either leave it as is or remove the entries from http://trojanedbinaries.com/security/dynamicdns altogether from `domain.txt` because of lots of junk inside

MikhailKasimov:
OK, I propose to leave junks in ```domain``` and move all other ones to ```dynamic```.

stamparm:
In such case I would leave everything as is

MikhailKasimov:
OK, gonna close these two PRs.

PR #12219, opened 2020-11-12
Create thrower.txt

MikhailKasimov:


stamparm:
Will merge, though, I believe that these two cases are unrelated (2011 report for the VBS version)

MikhailKasimov:
"I believe that these two cases are unrelated (2011 report for the VBS version)" <-- possibly. I've never met this panel, so I just took its name. Thanks for merging!

PR #12440, opened 2020-11-21
Create vbvgrabber.txt

MikhailKasimov:


stamparm:
These (e.g. atsengine) are all old stuff used by banking malware. Also, `rbcgrab` seem to me like a custom naming for a panel.

Anyhow, as somebody mentioned in that Twitter thread, `zloader` should be a better naming

MikhailKasimov:
Have moved to ```zloader``` : https://github.com/stamparm/maltrail/pull/12472/commits/d017d44526a4540335b4cc1ab484dc0c8d13dbc7

PR #12554, opened 2020-11-25
Update zloader.txt

MikhailKasimov:


stamparm:
AFAICS, majority are compromised sites, while reports do not state clearly what paths are used

PR #12986, opened 2020-12-13
Update js_script.txt

MikhailKasimov:
See Figure 7 in article.


stamparm:
There are also regular frameworks which use this same name (e.g. http://www.injectjs.com/)

PR #13351, opened 2020-12-28
Update cobaltstrike.txt

MikhailKasimov:


stamparm:
Slightly skeptic regarding `gov-hr-no-reply.org`, but it doesn't look benign ATM. Will merge

PR #13481, opened 2021-01-04
Update masslogger.txt

MikhailKasimov:


stamparm:
Not a fan of these kind of IoCs, where shared server names (e.g. `serverXXX.web-hosting.com`) are used. In months or so they will be used for legit purposes.

Thus, I'll merge this, but please don't mind if I remove/reject in future

MikhailKasimov:
OK

PR #13485, opened 2021-01-04
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
This was already inside and I removed it months ago (AFAIR). Thing is that this domain is causing lots of noise because it is used as a regular service from appliances, while I don't see that you can use it the same way as used in case of other dynamic domain providers

PR #13514, opened 2021-01-05
Update crypto_mining.txt

MikhailKasimov:


stamparm:
`outlookcalendar.accesscam.org` is malware -> `generic.txt`

`crypto_mining.txt` should be used only for regular crypto mining services

MikhailKasimov:
"outlookcalendar.accesscam.org is malware -> generic.txt" -- done.

PR #13948, opened 2021-01-21
Update settings.py

MikhailKasimov:
https://blog.netlab.360.com/not-really-new-pyhton-ddos-bot-n3cr0m0rph-necromorph/ --> ```/keksec.x88_64```


stamparm:
`/x88` is a mistypo. here should go common ones

PR #14295, opened 2021-02-03
Update settings.py

MikhailKasimov:
https://github.com/stamparm/maltrail/pull/14280/commits/1bdc5af23746c95301da8855ac148dca3f84a8c5 --> ```ktn.pp-c```


stamparm:
we had similar discussion. will merge, but only those "repeating" patterns should go here

PR #14716, opened 2021-02-19
Update web_shells.txt

MikhailKasimov:


stamparm:
This is for generic detection of web shells on servers (not for specific trails pointing to compromised sites)

PR #14895, opened 2021-02-26
Update android_generic.txt

MikhailKasimov:


stamparm:
Not sure if smart to merge `/gallery.apk` and similar (also, those previous similar trails)

PR #15207, opened 2021-03-10
latest DataPlane.org feeds added

jtkristoff:
Hello,

There have been some new DataPlane.org feeds since the last pull request.  I've added them for your convenience.  It looks like the original independent feed parsers are no longer used as before.  All the feeds are the same format except proto41.txt has two time stamp fields, a "firstseen" and "lastseen".

John

### DIFF
--- a/trails/feeds/dataplane.py
+++ b/trails/feeds/dataplane.py
@@ -13,7 +13,7 @@ __reference__ = "dataplane.org"
 
 def fetch():
     retval = {}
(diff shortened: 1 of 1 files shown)


stamparm:
Hi John.

Thank you for your update. Now, I have a potential issue about the usability of `proto41` inside the Maltrail. Not sure that communication attempts from `open IPv4 protocol 41 relays` could be marked as suspicious/malicious. My impression (maybe I am wrong) is that those same IPs appearing on that list could be just a result of the misconfiguration/old network equipment, while client contacting from that same NAT-ed network should most probably be a valid user.

I am opened to discussion.

BR

jtkristoff:
Yes, that is exactly right and in fact at least a few months ago, as far as I know, was never publicly realized until some related research work of mine is about to be published in PAM 2021 later this month.  So as of now they may not be actively abused.

If that means rejecting that feed then I'll leave that up to you.  If there is a way for the feed to be used for low impact, reporting purposes only that would be ideal since that is what I recommend for all the feeds anyway.

stamparm:
so, let me be a bad reviewer :D. if the situation changes, we could continue the discussion regarding the `proto41.txt`

PR #15622, opened 2021-03-29
Create babydraco.txt

MikhailKasimov:
```/owa/auth/babydraco.aspx``` <-- ```On Mar. 26, Shadowserver saw an attempt to install a new type of backdoor in compromised Exchange Servers, and with each hacked host it installed the backdoor in the same place: “/owa/auth/babydraco.aspx.”```


stamparm:
1) `babydraco.aspx` is a web shell, not an indicator of malware infection
2) `krebsonsecurity` domain is used for a miner

MikhailKasimov:
Resounded in https://github.com/stamparm/maltrail/pull/15637 and https://github.com/stamparm/maltrail/pull/15638

PR #15699, opened 2021-04-01
Create empire.txt

MikhailKasimov:


stamparm:
`generic.txt`

MikhailKasimov:
Idea is the same to ```/malicious/covenant.txt```. When merged, Wiki-page "Specific detections" also gonna be updated.

stamparm:
But what is "Empire" in this case?

stamparm:
If it is Powershell Empire, then it is strange to have only one trail for something that is for years around

MikhailKasimov:
> If it is Powershell Empire,

Looks, it is https://github.com/BC-SECURITY/Empire in fact.

>  then it is strange to have only one trail for something that is for years around

We started to track ```/malicious/covenant.txt``` also with few of trails only...

PR #16316, opened 2021-05-01
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
The same pull request came years ago. Rejected for the following: this is not regular DDNS. It is a home user service for sharing synology shares. Won't include it here as it will generate tons of events

PR #16328, opened 2021-05-02
Create apt_39.txt

MikhailKasimov:
Figure 4 and Figure 5 from Securelist article.


stamparm:
https://malpedia.caad.fkie.fraunhofer.de/actor/apt39

PR #17406, opened 2021-06-27
Create mielit.txt

MikhailKasimov:


stamparm:
generic?

MikhailKasimov:
It is temporary name for trail. I wanna wait, if some other samples are met. Too specific port to be ```generic``` one.

MikhailKasimov:
UPD: All samples from https://www.virustotal.com/gui/domain/spazioprograms.no-ip.biz/relations are ```mielit\mesub```.

PR #17823, opened 2021-07-25
Update dynamic_domain.txt

MikhailKasimov:


stamparm:
these don't look like dynamic domains :|
just as an example: https://www.virustotal.com/gui/domain/dyndns.org/relations (look at the variety of subdomains, variety of IP addresses. also, it has a proper registration page)

MikhailKasimov:
-> ```domain.txt```?

stamparm:
these look like some random internet domains. i don't see anything interesting (i.e. malicious) regarding them

MikhailKasimov:
The main deal was here: https://github.com/stamparm/maltrail/pull/17826/files <-- mix of domains and dyn-domains. Those domains, which have invariantive IPs went to ```domain.txt``` (https://github.com/stamparm/maltrail/pull/17824/files). Those ones, that have moretheless variantive IPs --> to ```dynamic_domains.txt```. So, let's decide where to keep these current domains...

stamparm:
this https://github.com/stamparm/maltrail/pull/17826/files seems to me that malware authors mixed legit and dynamic-domains to mask the proper domain. thus, all related main domains should not go anywhere

MikhailKasimov:
OK

PR #18170, opened 2021-08-06
Update worst_asns.txt

MikhailKasimov:


stamparm:
this is a generic list, not prone to changes (highly subjective list). we had similar discussion before

PR #19030, opened 2021-12-13
Update settings.py

MikhailKasimov:
Update due to https://twitter.com/HaboubiAnis/status/1470415975220756486 and https://github.com/SigmaHQ/sigma/blob/master/rules/web/web_cve_2021_44228_log4j_fields.yml info.


stamparm:
Please, now pause with all these log4j updates

PR #19032, opened 2021-12-23
dynamic domains - add sitelutions dyndns list

wesinator:
https://www.sitelutions.com/info/slredirection

Associated with historical malware, e.g. https://otx.alienvault.com/indicator/hostname/kita.myfw.us


MikhailKasimov:
Currently ```kwikto.com``` is missed only. All others are in ```domain.txt``` (historically).

stamparm:
this is redirection service, not "dynamic domain" (i.e. DDNS)

wesinator:
> this is redirection service, not "dynamic domain" (i.e. DDNS)

interesting - my understanding was that the domains can function like ddns, pointing to any IP address configured.

PR #19036, opened 2021-12-31
Update agenttesla.txt

MikhailKasimov:


stamparm:
Looks more like compromised credentials are used by operators. Domain is valid >2 years, it has some Google presence, etc.

PR #19040, opened 2022-01-12
Update settings.py

MikhailKasimov:
https://twitter.com/wugeej/status/1481158216390426628 <-- Looks like even not necessary to run ```sh /tmp/``` here.


stamparm:
no need to include everything inside. we should cover the majority with as low number of rules are possible

PR #19046, opened 2022-02-06
Add more mass scanner IPs

t3chn0m4g3:
IP findings are based off the original maltrail project work, extended by findings in my personal honeypot logs and enriched through search engine searches using sublist3r (passive only).
Added a CIDR based list `mass_scanner_cidr.txt` intentionally using only the nearest /24 to cut down on number of blackhole routes / firewall rules.


MikhailKasimov:
@t3chn0m4g3 Thank you for updating!

t3chn0m4g3:
Sure thing, thank you for maltrail.

Issue #11332, opened 2020-10-04
How are parked sites worked out?


youradds:
Hi,

I'm just wondering how parked sites are worked out? Most of the ones I've checked are indeed parked - but I just came across this one:

"2020-10-04 07:28:12.065448" x.com 127.0.0.53 53 127.0.0.1 50268 UDP DNS (worthingtonseniorliving).com "parked site (suspicious)" (heuristic)

It seems live though? http://worthingtonseniorliving.com/ 

TIA

Andy

youradds:
You know what - scrap that! That IS a holding page! It just looked a bit more fancy than a standard one. I just had another look at it, and can see its a holding page. Sorry about that ;)

stamparm:
It is based on a [list](https://github.com/stamparm/maltrail/blob/master/trails/static/suspicious/parking_site.txt) of known parking IPs. 

If you go to https://www.virustotal.com/gui/ip-address/208.91.197.27/relations, you'll see a huge number of same looking sites. Thus, this indeed looks like parked
![1](https://user-images.githubusercontent.com/921555/95059009-9bc58000-06f8-11eb-90f8-8ad52c452966.png)
![2](https://user-images.githubusercontent.com/921555/95059013-9c5e1680-06f8-11eb-8ba1-cae2baf19d5e.png)

youradds:
Thank you for the explanation :)

PR #19057, opened 2022-03-25
Add {buy, sell, sell2}.internettraffic.com in parking_site.txt

PeterDaveHello:
Found this when digging parking domain related issue like: #11332, maybe `worthingtonseniorliving.com` changed its parking service, looks like the new one was not recorded here.

image


stamparm:
Only `52.128.23.153` seems to be active and actually being used as "parking site"

PR #19083, opened 2022-05-12
Update plugx.txt

ilumitr:
Update recent PlugX IoC


MikhailKasimov:
Dup of https://github.com/stamparm/maltrail/blob/master/trails/static/malware/apt_mustangpanda.txt#L129

PR #19093, opened 2022-06-07
Remove fin7 false positives

comawill:
The removed domains are part of Wire's production infrastructure
(*.wire.com) and staging/testing infrastructure (*.zinfra.io /
*.wire.link)

The related fin7 sample was a modified Wire client.

This probably also resolves https://github.com/stamparm/aux/issues/5.


MikhailKasimov:
Looks, you are right. Thank you for reporting!

PR #19133, opened 2022-11-29
Update dynamic_domain.txt

MikhailKasimov:
@stamparm Please, verify the list. I'm not sure couple of records are related to dyn-dns. If OK to move to ```suspicious\domain.txt``` -- will move.


MikhailKasimov:
No related to dyndns.

PR #19134, opened 2022-12-02
Update apt_mustangpanda.txt

GossiTheDog:


MikhailKasimov:
Duplication of https://github.com/stamparm/maltrail/blob/master/trails/static/malware/plugx.txt#L331-L336

PR #19140, opened 2023-01-23
Update whitelist.txt

hbednar:
These are all false positives and being detected as suspicious. Except, exploit-db which is being detected as malware though it is infarct a legitimate website.

### DIFF
--- a/misc/whitelist.txt
+++ b/misc/whitelist.txt
@@ -32,7 +32,6 @@ a-msedge.net
 angsrvr.com
 anycastb.com
 anti-virus.by
-api.ipify.org
 azure.com
 b4b.hr
 baidu.com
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
> > `myip.ms` and `api.ipify.org` -- these are ipinfo services. Won't be whitelisted.
> 
> Is there a way to whitelist these services without manually adding them each time.

Yes: just put ```#``` sign in front of trail: e.g. ```# myip.ms```, save trail and restart the sensor.

Also you can apply ```USER_WHITELIST``` variable in ```/maltrail.conf``` file: https://github.com/stamparm/maltrail/blob/master/maltrail.conf#L121-L122

> 
> > ` exploit-db which is being detected as malware` <-- by what trail it being detected as malware is? Because I can't find it, except one single mentioning in `# Reference:` section of `/malware/volk.txt` base: image . But this doesn't trigger the detection.
> 
> 

Ah, it is data from feed, OK... Anyway, already whitelisted.

PR #19141, opened 2023-01-24
Update whitelist.txt

hbednar:
Several false positives that are being detected as suspicious, some of these are torrent trackers.

### DIFF
--- a/misc/whitelist.txt
+++ b/misc/whitelist.txt
@@ -207,6 +207,26 @@ zextras.com
 zillya.com
 zillya.ua
 zillyaoem.com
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
Currently thinking... On the one hand -- torrent trackers, on another -- they are on root domains, which are mostly seen in malware trails...

hbednar:
I know torrenting doesn't always have a good reputation but, i am not aware of any of these domains being associated with malicious activity. Also, i dont think any of them appear in any of the feeds.

stamparm:
Valid case. Thus, merged

PR #19230, opened 2023-11-23
Add blacklists

akiuni:
This PR aims at improving the FAIL2BAN regexp which can be considered too generic. I've added two blacklists (IN and OUT) with an easier way to configure.
From my point of view the FAIL2BAN regexp could be removed and replaced by `OUT_BLACKLIST` and `IN_BLACKLIST`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.py[cod]
 *~
+Pipfile*
--- a/README.md
+++ b/README.md
@@ -251,22 +251,25 @@ for ip in $(curl http://127.0.0.1:8338/fail2ban 2>/dev/null | grep -P '^[0-9.]+$
 iptables -I INPUT -m set --match-set maltrail src -j DROP
 ```
 
-Options `OUT_BLACKLIST` and `IN_BLACKLIST` are slightly different than `FAIL2BAN_REGEX`. They allow to build regular expressions to apply on one field. For each rule, the syntax is : `<field> <control> <regexp>` where :
+Option `BLACKLIST` allow to build regular expressions to apply on one field. For each rule, the syntax is : `<field> <control> <regexp>` where :
 * `field` indicates the field to compage, it can be: `src_ip`,`src_port`,`dst_ip`,`dst_port`,`protocol`,`type`,`trail` or `filter`.
 * `control` can be either `~` for *matches* or `!~` for *doesn't match*
 * `regexp` is the regular expression to apply to the field.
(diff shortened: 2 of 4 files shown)


MikhailKasimov:
@stamparm looks interesting

akiuni:
wait there is a bug... I'm fixing it

akiuni:
Done  :-)

I've created a single commit for the bug I've fixed because it is out of the scope of this PR and may be already fixed in another branch : https://github.com/stamparm/maltrail/pull/19230/commits/7a49e8f6880ac19cbcb3d4c8bc20c319745da0c7

I've added a subpath parameter so as to give blacklist names: https://github.com/stamparm/maltrail/pull/19230/commits/4d9137db98b35babc8763647b2ce88daa7bd4688 , however, this is just a proposal.

So with this PR, users will be able to build friendly named blacklist, depending on their needs.

I hope you will appreciate my PR, I can contribute to your project to fix bugs and maintain this part if you want.

Best regards
Julien

MikhailKasimov:
```So with this PR, users will be able to build friendly named blacklist, depending on their needs.``` <-- question here: how\in which way these blacklists could be passed to some blocking mechanism (e.g. iptables, pf, etc)?

**Comment by akiuni** [comment]
> `So with this PR, users will be able to build friendly named blacklist, depending on their needs.` <-- question here: how\in which way these blacklists could be passed to some blocking mechanism (e.g. iptables, pf, etc)?

I've explained it in the README.md ( here ->  https://github.com/stamparm/maltrail/pull/19230/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) 

For example if you create a blacklist named `TEST` like that : 
```
BLACKLIST_TEST
   (rules)...
```

You can get the maching ip addresses with a curl on [http://x.x.x.x:8338/blacklist/test]()
(the same way like you did with fail2ban)

MikhailKasimov:
Thnx, that's ok!

stamparm:
been looking into code:

1) Maltrail should stay python2 compatible. that means that stuff like:
`bl_name = f"_{params['subpath'].split('/')[0].upper()}"`
and
`splitpath = path.split('/', maxsplit=1)`
will break its backward compatibility

2) adding ` == n :` at the end of the line is not really pythonic (NOTE: it should be ` == n:`)

3) there should be spaces between operator commas at `for f,n,r in bl:` and `['','','','src_ip','src_port','dst_ip','dst_port','protocol','type','trail','filter']`

akiuni:
Thank you @stamparm for your review 

You are absolutely right, I didn't check the linter.. Sorry, I review my code.

About the *non-pythonic * test, I'll see if I can make it with a xnor gate if you prefer

I rename the PR as WIP in the meanwhile

akiuni:
Done

MikhailKasimov:
@stamparm And?... Yes or no?

PR #19234, opened 2023-12-05
Update whitelist.txt

hbednar:
This is a github ip and part of their range 185.199.108.0/22 and it is appearing as a mass scanner false positive.

Full ip list 
https://api.github.com/meta


MikhailKasimov:
Thank you! Plus fix: https://github.com/stamparm/maltrail/commit/7ddd9cd70d5ede79f58231229b512b2048d3149d

PR #19240, opened 2024-01-17
Remove improper aliases for teamspy

HSZemi:
Judging by the commit history, it seems like the aliases "phichichi" and "socks5systemz" were taken from this Tweet:
https://twitter.com/g0njxa/status/1701212547305607283
image

However, the tweet seems to instead indicate that the given IP addresses are Socks5 Systemz systems that are being used by TeamSpy. It does not suggest that Socks5 Systemz is the same as TeamSpy.

Also, Pichichi/phichichi appears to be just the title on the displayed website, not the name of a malware family.

See also:
- https://malpedia.caad.fkie.fraunhofer.de/details/win.teamspy
- https://malpedia.caad.fkie.fraunhofer.de/details/win.socks5_systemz

### DIFF
--- a/trails/static/malware/teamspy.txt
+++ b/trails/static/malware/teamspy.txt
@@ -1,8 +1,6 @@
 # Copyright (c) 2014-2023 Maltrail developers (https://github.com/stamparm/maltrail/)
 # See the file 'LICENSE' for copying permission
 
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
This simplified the search, when you can't remind what is what. "Thanks" for that...

PR #19279, opened 2024-10-17
Update vidar.txt to remove false positive

olt:
This Hetzer IP belongs to a new server since ~2024-03-15 and is now a false positive. The IP is also no longer listed on https://threatfox.abuse.ch/browse/malware/win.vidar/ (the source for this entry from 2024-02-12).


MikhailKasimov:
Thank you for alerting!

PR #19298, opened 2025-02-12
fix: remove EOL DNSBL rbl.lugh.ch

weeheavy:
The lugh RBL is EOL, this PR removes it from `misc/whitelist.txt`


MikhailKasimov:
Hello!

Reason, please?

weeheavy:
I am the owner of that service and decomissioned it in late 2024, which can also be seen from its DNS TXT response.

MikhailKasimov:
OK, got that. Thank you for the info!

PR #19305, opened 2025-03-21
Update blackholemonster.py

chars6:
Fix garbled return data.


stamparm:
they have put Cloudflare in between

BlackHoleMonster:
Sorry, had to rework it, now works.

https://blackhole.s-e-r-v-e-r.pw/blackhole
https://blackhole.s-e-r-v-e-r.pw/blackhole-today

PR #19306, opened 2025-03-21
Update blackholemonster.py

chars6:
having problems with database (appending error messages to the end of gzip stream)


MikhailKasimov:
OK, feed is back: https://github.com/stamparm/maltrail/commit/a55649c54e313a9f2f8ec89f23b0b1a75edb6f13

PR #19312, opened 2025-05-01
Update README.md

CharlesCNorton:
- “As of static entries” → “As **for** static entries”
- “Using of pcapy”      → “Use of pcapy”
- “in general case”     → “in **the** general case”
- “in bytes of percentage” → “in bytes **or** percentage”
- “compage”             → “compare”
- “there is no need install” → “there is no need **to** install”
- “Mass scans is”       → “Mass scans **are**”
- “It the following screenshot” → “**In** the following screenshot”


MikhailKasimov:
Thank you!

PR #19327, opened 2025-07-18
added onyphe.net IP addresses

mlodic:
same as the previous PR but for Onyphe http://dixon.probe.onyphe.net/ thanks :)


mlodic:
@MikhailKasimov ready to be added. Please let me know if the format is fine, I added the ranges in the specific cidr list and all the related addresses in the other file. Thanks

PR #19335, opened 2025-07-25
feat: Improve Dockerfile

simonszu:
I have restructured the Dockerfile. 
- Moved all Docker related stuff to a subfolder
- Modified the README to resemble the new Docker subfolder
- Restructured the Dockerfile for better layering and removing unneeded cache
- Introduced tini as a small process supervisor so that two processes (maltrail itself and cron for trail update) are running in the container 
- Log messages are redirected to STDOUT which is the usual way of getting logs from a Docker container

As a sidenote it is currently not possible to start the Docker container in sensor mode, since the python command is now hidden in a start.sh-script. There are ways to solve this with ENV variables and envsubst, but this is something i'll tackle in the future.


MikhailKasimov:
@ki9us FYI.

PR #19540, opened 2026-03-26
Add backend support for “past occurrences” count per event row

### LINKED ISSUE #19235: [Feature Request] Show Number Of Past Entries
**Is your feature request related to a problem? Please describe.**
You can only see results for that day and cannot see if one of the events has been seen before in the past.
Please let me know if this is possible and i have missed something.

**Describe the solution you'd like**
It would be great if there was another column next to events showing the number of times this event has been seen before.

**Describe alternatives you've considered**

**Additional context**

huyhoang171106:
## Summary

Extend the reporting endpoint that returns event rows so each row includes an additional numeric field representing how many times the same event was observed **before the selected day/time window**. This must be computed server-side so the UI can display it without additional round trips.

## Files changed

- `server.py` (modified)

## Testing

Closes #19235

### DIFF
--- a/server.py
+++ b/server.py
@@ -11,6 +11,9 @@ import sys
 
 sys.dont_write_bytecode = True
 
+import collections
+import csv
+import glob
 import optparse
 import os
 import platform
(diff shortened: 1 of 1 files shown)


stamparm:
This landed differently in the meantime: there's a condensed observable store (meta.sqlite) that keeps first-seen/last-seen/count per domain or IP, and the detail drawer shows it as a "network memory" line. Doing the count over the event log per row was the expensive part this avoids. Thanks anyway.

PR #19541, opened 2026-03-26
Add strict domain classification and non-domain indicator tracking for static trails

### LINKED ISSUE #19161: [Feature Request] non-domain indicators parser for /maltrail-malware-domains.txt
Inspired by: https://github.com/stamparm/aux/issues/8

```
[i] Target: https://raw.githubusercontent.com/stamparm/aux/master/maltrail-malware-domains.txt
[✓] Status: Retrieval successful
[✓] Parsed 290973 exact domains and 0 ABP-style domains (ignored 13 non-domain entries)
```

To have a filter\parser for non-domain indicators when collecting domains from ```maltrail static trails``` to ```/maltrail-malware-domains.txt```:

```maltrail static trails --> [non-domain indicators] --> /maltrail-malware-domains.txt```

In case of ```non-domain indicator != 0```, to send emails to MS and MK (e.g. ...| ```mail -s "non-domain entries found" email@email.email```).

This should speed up correction, if junk/orphan records are going to get incorporate into ```/maltrail-malware-domains.txt```.

huyhoang171106:
## Summary

Static trail parsing currently normalizes many values into a domain-like form without validating whether the final token is actually a valid FQDN. This allows junk/orphan indicators (hashes, malformed hostnames, accidental tokens) to be treated as domains by downstream exporters.

## Files changed

- `trails/static/__init__.py` (modified)

## Testing

Closes #19161

### DIFF
--- a/trails/static/__init__.py
+++ b/trails/static/__init__.py
@@ -14,8 +14,27 @@ from thirdparty import six
 
 __url__ = "(static)"
 
+_non_domains = set()
+
+def _is_domain(value):
+    if not value or '.' not in value or len(value) > 255:
+        return False
+
+    labels = value.split('.')
+    if re.search(r"\A\d+\Z", labels[-1]):
+        return False
+
+    for label in labels:
+        if not label or len(label) > 63 or label[0] == '-' or label[-1] == '-':
+            return False
+        if not re.search(r"\A[a-zA-Z0-9-]+\Z", label):
+            return False
+
+    return True
+
 def fetch():
     retval = {}
+    _non_domains.clear()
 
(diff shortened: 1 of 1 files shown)


stamparm:
Thanks for looking at this, but this would quietly cut a quarter of the static set. I ran the check over trails/static as it stands: of the 1.83M entries that aren't plain IPv4, 457,774 fail `_is_domain()` - 287k of them ip:port and IPv6 entries, 169k URL/path trails, plus the CIDR ranges and the wildcard regex patterns. Those are all deliberate trail shapes, not junk, and dropping them is the worst kind of bug here (everything still starts, it just stops detecting). If the goal is catching typos in the domain files, a validation script that *reports* suspicious lines is welcome - but the loader shouldn't silently discard them.

PR #19543, opened 2026-03-26
Introduce HOPE UI-based layout shell while preserving Maltrail data hooks

### LINKED ISSUE #19049: A new design
Hello,
Is it possible to create a new design with the HOPE UI template ?
https://hopeui.iqonic.design/

huyhoang171106:
## Summary

The current UI is tightly coupled to existing element IDs/classes used by `html/js/main.js`. To avoid breaking event bindings and data rendering, update the HTML structure to a modern HOPE UI-style shell (top bar, filter card, events card, details panel) but keep all functional IDs and input names unchanged. Add a theme switch control in the header and load a new HOPE UI override stylesheet after existing styles so the redesign is opt-in and non-breaking.

## Files changed

- `html/index.html` (modified)

## Testing

Closes #19049

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -9,12 +9,13 @@
         <link rel="stylesheet" type="text/css" href="css/thirdparty.min.css">
         <link rel="stylesheet" type="text/css" href="css/main.css?id=222">
         <link rel="stylesheet" type="text/css" href="css/media.css">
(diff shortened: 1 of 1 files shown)


stamparm:
The UI was rewritten since this was opened - new layout, themes, charts, drawer, no build step - so this shell doesn't apply any more. Thanks for the idea though, the design nudge did land (see #19049).

PR #19586, opened 2026-08-02
Remove false-positive blacklist entry for my domain

mrpiesel:
I host a small video chat app, and i noticed that my primary domain (along with one of its RTC server domains) is listed on your blacklist.

This appears to be a false positive: the domain is not present on any other blacklist like virustotal, and the referenced X post seems to have been removed. I don't know when and why it was added, as my project has been inactive for the past few months. I was planning to resume development, then I noticed that my dns blockers are blocking the domain.

### DIFF
--- a/trails/static/malware/elf_mirai.txt
+++ b/trails/static/malware/elf_mirai.txt
@@ -58207,21 +58207,6 @@ http://93.123.109.160
 shinju-networks.net
 cnc.shinju-networks.net
 
-# Reference: https://x.com/redrabytes/status/1849337367863189820
-# CLASS_0_HASH-HOST=82f94b386bd4d7dc9f04a19553c244a9
-
-141.98.10.116:53648
-154.213.187.58:48920
-notcord.vip
-novo.geek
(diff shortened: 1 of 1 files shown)


MikhailKasimov:
OK, please, explain this: https://www.virustotal.com/gui/domain/novoline.top/relations

image

mrpiesel:
> OK, please, explain this: https://www.virustotal.com/gui/domain/novoline.top/relations
> image

Didn't see this. I only did a quick check for detection and community tabs on VT to see, if there is fairly recent malicious activity on a few of those domains. Anyway, those are not mine. Only domain from this list that i own is the notcord.vip and i'm looking to get it removed. Maybe it was added cause in the past a few of my rtc servers was hosted by external people and they did something?

MikhailKasimov:
OK, would get fixed for now: https://github.com/stamparm/maltrail/commit/04855b95d4fe79e7e677653d669cd564211de0bd

mrpiesel:
Thanks

PR #19594, opened 2026-08-09
Add IPWhois.net community blacklist feed

brkicsivan-wq:
Adds a fetcher for the IPWhois.net community blacklist feed.

- Feed: https://bl.ipwhois.net/feed.txt (plain text, one IPv4 per line, `#` comments)
- Content: first-party reports only - IPs reported by users via the API, fail2ban/CSF integrations and our own honeypots. Third-party feeds that IPWhois.net mirrors on its website (blocklist.de, CI Army, IPsum, ...) are deliberately excluded from this file, so ingesting it does not create circular data.
- Criteria: confidence >= 50, seen within the last 90 days. Currently ~1,000 entries, regenerated every 30 minutes.
- Free to use, no key or signup. Docs: https://ipwhois.net/blacklist/docs

Fetcher follows the same pattern as greensnow.py. `__check__` matches the fixed header comment of the feed.

Disclosure: I run IPWhois.net.

### DIFF
--- a/trails/feeds/ipwhois.py
+++ b/trails/feeds/ipwhois.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2014-2026 Maltrail developers (https://github.com/stamparm/maltrail/)
+See the file 'LICENSE' for copying permission
+"""
+
+from core.common import retrieve_content
+from core.settings import NAME
+
+__url__ = "https://bl.ipwhois.net/feed.txt"
+__check__ = "IPWhois.net Blacklist"
+__info__ = "known attacker"
+__reference__ = "ipwhois.net"
+
+def fetch():
+    retval = {}
+    content = retrieve_content(__url__, headers={"User-agent": NAME})
+
+    if __check__ in content:
+        for line in content.split('\n'):
+            line = line.strip()
+            if not line or line.startswith('#') or '.' not in line:
+                continue
+            retval[line] = (__info__, __reference__)
+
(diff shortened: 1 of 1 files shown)


stamparm:
Ran it against the real loader: 882 IPs, feed is live and the `__check__` guard matches, and "known attacker" is the right label for honeypot/fail2ban reports. Merging - thanks!

stamparm:
Reverted this one too - and the reason is on me rather than on you, since I merged it after checking only that the module loads and the feed answers.

Putting numbers on it afterwards: established attacker lists corroborate each other heavily - blocklist.de and greensnow share 29% of the smaller list, blocklist.de and abuseipdb 29%, greensnow and abuseipdb 56%. bl.ipwhois.net shares 0.3% with greensnow, 3.7% with blocklist.de and 21% with abuseipdb. Of its 882 addresses, 20 appear anywhere in Maltrail's current 1.6M-entry set, and those are labelled mass scanner rather than attacker.

SSH brute-force reporting is heavily redundant across lists, so near-zero agreement with peers is hard to read as novel coverage. Add that ipwhois.net goes back to 2004 but the blacklist under it only shows up in the Wayback Machine from April 2026, with self-published metrics and an open report endpoint, and it doesn't clear the bar for a source we ship to every install.

I've written that bar down in .github/CONTRIBUTING.md so it isn't a matter of who reviews it. If the list matures - a published false-positive process, and overlap with the established lists that says it's finding the same attackers - it's worth revisiting.

PR #19605, opened 2026-08-17
Add ThreatCluster feed

Jam0k:
Adds the ThreatCluster public IOC feed (high-confidence malicious domains and IPs, validated, 30 day window, hourly refresh, no auth). Free direct downloads, details at https://threatcluster.io/iocs. Tested fetch() locally against the live endpoints.

### DIFF
--- a/trails/feeds/threatcluster.py
+++ b/trails/feeds/threatcluster.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+
+"""
+Copyright (c) 2014-2026 Maltrail developers (https://github.com/stamparm/maltrail/)
+See the file 'LICENSE' for copying permission
+"""
+
+from core.common import retrieve_content
+
+__url__ = "https://threatcluster.io/api/iocs/public/domains.txt#https://threatcluster.io/api/iocs/public/ips.txt"
+__check__ = "ThreatCluster public IOC feed"
+__info__ = "malicious"
+__reference__ = "threatcluster.io"
+
+def fetch():
+    retval = {}
+
+    for url in __url__.split('#'):
+        content = retrieve_content(url)
+
+        if __check__ in content:
+            for line in content.split('\n'):
+                line = line.strip()
+                if not line or line.startswith('#'):
+                    continue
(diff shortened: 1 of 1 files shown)


stamparm:
Checked it end to end - both URLs answer, the guard string is there, and the loader gets 35 entries (20 domains + 15 IPs) which look like real phishing/C2 infrastructure. Small window, but high signal. Merging.

One thing to pass back upstream: a few of the IPv6 entries are network prefixes (`2605:6400::`, `2a0a:d683::`) rather than host addresses, so they'll only ever match that exact address here.

stamparm:
Reverted this - my mistake for merging it, and I owe you a straight explanation.

What I actually checked before merging was that the module loads, the URLs answer and the guard string is present. That is plumbing, not quality, and the quality does not hold up:

- `supportxmr.com` and `pool.supportxmr.com` are the largest public Monero mining pool. Maltrail already carries them, correctly, as `crypto mining (suspicious)` in trails/static/suspicious/crypto_mining.txt. They are in the feed as malicious.
- `sinbad.io` and `yomix.io` are mixers that were sanctioned and seized back in 2023, so they are not findings from the last 30 days either.
- Three of the fifteen "IPs" are bare IPv6 network prefixes (`2605:6400::`), which only ever match that exact address.

That pattern is what automated extraction from articles produces: a cryptojacking write-up names the pool the miner connects to, and the extractor cannot tell the pool from the malware. Four questionable entries out of twenty in a list published as high-confidence is too high for a set that ships to every Maltrail install.

Separately, `__url__` was two URLs glued together with a '#'. update.py prints that string to the operator during updates - trails/feeds/dataplane.py shows how a multi-URL feed is written here (a readable `__url__`, the real URLs in a tuple inside `fetch()`).

<strike>Happy to look again if the feed grows a published validation/false-positive process, or if the public list is narrowed to things that are unambiguously attacker-controlled.</strike> <- sorry man. these mistakes were TOO HUGE to regain the trust for this or similar feeds

Jam0k:
Appreciate your time anyway. Thanks.

